PR #9, opened 2013-10-24
SLING-3179 - Implement solution to the Authentication Handler Credential...

tmaret:
... Validation Problem
- Add ResourceResolverFactory API for pre-authenticated users
- Extract JcrResourceProviderFactoryActivator which starts JcrResourceProviderFactory service
- Access 'usingBundle' via a ServiceFactory


kwin:
@tmaret Can you add a comment why you closed this, because it does not seem this has been merged in the meantime (https://issues.apache.org/jira/browse/SLING-3179).

tmaret:
@kwin indeed, thanks for noticing! The PR has not been merged, however the patch is still available in the JIRA issue if someone looks at it later
https://issues.apache.org/jira/secure/attachment/12610074/SLING-3179.diff

PR #10, opened 2014-01-29
SlING-3343 Sling-Junit-Core Method Name in Test selctor is not honoured

sufyanharoon:
Now method name is honoured while executing tests from browser.
Method name is added as part of test execution path i n junit and
slingjunit servlets which was not getting included.


rombert:

PR #14, opened 2014-04-22
SLING-3459, do not log exceptions which are rethrown

kwin:
Do not log exceptions which are rethrown, to prevent stacktraces from appearing twice in the log. The original bug report is at https://issues.apache.org/jira/browse/SLING-3459.


justinedelson:
I've merged this pull request, but there seems to be an infra problem which is preventing this pull request from being updated.

PR #15, opened 2014-04-23
SLING-3510 Check for null classUrls and warn when no @Model classes found

cpilsworth:
When a non-existent package is added to Sling-Model-Packages, then an NPE is thrown as there are no matching classes found. I guess the same would happen for existing packages that contain no Model classes. Check for classUrls and warn, then continue if null.


justinedelson:
Patch has been applied. PR status may not be updated due to apache infra issues.

cpilsworth:
Closing PR as this has been merged as patch.

PR #16, opened 2014-05-08
SLING-3547, add test for numerical defaults (both boolean and long,

kwin:
test exposing the wrong behaviour of numerical defaults in Sling Models


justinedelson:
I've applied a slightly modified version of this patch. Thanks.

PR #18, opened 2014-06-04
SLING-3612, add m2 project configuration for packaging bundle (secondary to tycho)

kwin:
Add project configurator which adds the Sling Bundle Facets automatically for all Maven Projects with packaging "bundle"


rombert:
Patch applied in [r1604646](http://svn.apache.org/viewvc?view=revision&revision=1604646), thanks for taking the time to submit this.

PR #19, opened 2014-06-06
SLING-3651, improved error message for conversion to content project

kwin:
always enable the conversion action (to be able to see the error
message)


kwin:
Closing because it was merged.

PR #21, opened 2014-06-24
SLING-3695 - TenantProvider throws NPE when listing tenants root tenant ...

tmaret:
...resource does not exist
- fix NPE
- add unit test


tmaret:
Patch has been applied

PR #22, opened 2014-07-04
SLING-3737 - Instance Sling Identifier may be randomly reset on restart

tmaret:
- add unit test to cover the use case
- read the file completely using org.apache.commons.io.FileUtils
- fix the writeSlingId method not taking the passed identifier into account


ieb:
Does the unit test reproduce the failure pattern ?

tmaret:
Fix has been integrated, closing.

PR #23, opened 2014-07-21
SLING-3776 JsonRendererServlet should provide a way to list children as ...

visiongeist:
Adding an "array" selector will print the children of the node under the key "_children" and adds its key as a "_name" attribute. The wordings would need to be improved.

I did the implementation of it only in the servlet, some parts may be possible to be moved to JSONObject though imho I think it belongs rather to the servlet because it is for pure formatting.

### DIFF
--- a/bundles/commons/json/src/main/java/org/apache/sling/commons/json/JSONObject.java
+++ b/bundles/commons/json/src/main/java/org/apache/sling/commons/json/JSONObject.java
@@ -1192,7 +1192,7 @@ public class JSONObject {
      *  with <code>}</code>&nbsp;<small>(right brace)</small>.
      * @throws JSONException If the object contains an invalid number.
      */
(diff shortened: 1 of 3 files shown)


bdelacretaz:
Could you add automated tests? Ideally alongside the  https://svn.apache.org/repos/asf/sling/trunk/launchpad/integration-tests/src/main/java/org/apache/sling/launchpad/webapp/integrationtest/JsonRenderingTest.java integration test. The ../integration-tests/README.txt explains how to run those tests.

visiongeist:
I'm on it

rombert:

bdelacretaz:
The `.harray` selector is implemented and documented at https://sling.apache.org/documentation/bundles/rendering-content-default-get-servlets.html so I think this can be safely deleted.

PR #24, opened 2014-08-06
SLING-3709, add another exception-aware mechanism of instanciating Sling

kwin:
SLING-3709, add another way of instanciating a Sling Model which throws meaningful exceptions in case of errors


kwin:
Patch has been applied.

PR #28, opened 2014-10-07
SLING-4010, allow validator to set custom validation message

kwin:
https://issues.apache.org/jira/browse/SLING-4010


kwin:
Patch has been applied.

PR #29, opened 2014-10-08
SLING-4012, do not catch SlingValidationExceptions

kwin:


kwin:
Patch has been applied.

PR #30, opened 2014-10-09
SLING-4024 - Creates a new MockResourceResolverBuilder class

mcdan:
- This class is capable of creating a resource resolver based
  on JSON data that can be retrieved from sling's .json selector
- Includes test that cover its function
- Had to add org.apache.sling.commons.json to the pom to enable
  parsing


rombert:
@mcdan  - the Jira issue has been resolved, can you please close the pull request?

rombert:

PR #31, opened 2014-10-21
SLING-4079 - The LaunchpadConfigInstaller should skip over non-installable resources

raducotescu:
The attached commit skips over non-installable resources. All tests have run successfully after the change.


raducotescu:
The proposed patch was applied in commit c4c8b01303a4475e5dad196a3223a89d4be09480. Closing this PR.

PR #32, opened 2014-10-22
SLING-4056, support adapters on Sling Model Factory

kwin:
also additional test cases for the factory were provided


kwin:
Patch has been applied.

PR #34, opened 2014-11-12
SLING-4157, allow models in any location

kwin:


kwin:
Patch has been applied.

PR #35, opened 2014-11-12
SLING-4159, close resource resolver in finally

kwin:


kwin:
patch has been applied.

PR #36, opened 2014-11-13
Update and rename README.txt to README.md

stokito:


rombert:

PR #39, opened 2014-11-17
SLING-4138, refactor Validator interface

kwin:
refactor Validator interface to support
- acting on arbitrary data types (not only String)
- cross-checks for multiple properties
- array validation


kwin:
The changes have been applied.

PR #44, opened 2014-11-18
SLING-4182 - Cleanup Sightly Engine dependencies

raducotescu:
- reverted `commons-io` to 2.4 due to the incorrect metadata info from 2.0


fmeschbe:
I get your point. I suggest we actually drop down to commons-io 1.4. This still suffices the API and has the appropriate export (and aligns with what we have in Sling)

PR #49, opened 2014-11-27
SLING-4207 - Add notifications upon tenant changes

tmaret:
- Add API TenantEventListener that contains the methods invoked in case a tenant is added/removed or tenant properties are set/removed
- Support the TenantEventListener API in the implementation
- Increase minor package version


tmaret:
Closed this PR and opened a new PR in https://github.com/apache/sling/pull/49 that uses an event based approach as discussed on the list.

PR #50, opened 2014-11-28
SLING-4207 - Add notifications upon tenant changes

tmaret:
- Add API TenantConstants that defines the string constants to process the events.
- Send events when tenant is added/removed or tenant properties are set/removed from the implementation
- Increase minor package version


rombert:

PR #54, opened 2015-01-15
Sling 4258

santiagozky:
SLING-4258
-fix how the json servlet delivers dates with timezones.
-add test cases for some dates with different timezones


bundles/commons/json/src/main/java/org/apache/sling/commons/json/sling/JsonObjectCreator.java
```diff
@@ -119,10 +120,11 @@ private static JSONObject create(final Resource resource,
     /** Used to format date values */
     private static final Locale DATE_FORMAT_LOCALE = Locale.US;
 
-    private static final DateFormat CALENDAR_FORMAT = new SimpleDateFormat(ECMA_DATE_FORMAT, DATE_FORMAT_LOCALE);
 
     private static synchronized String format(final Calendar date) {
-        return CALENDAR_FORMAT.format(date.getTime());
+        DateFormat formatter = new SimpleDateFormat(ECMA_DATE_FORMAT, DATE_FORMAT_LOCALE);
```
kwin, line 125: The javadoc recommends to instantiate  the SimpleDateFormat per thread. You do that, which is good. But now we can get rid of the synchronized! That was only necessary previously because SimpleDateFormat was shared.

rombert:
@santiagozky  - the Jira issue has been resolved, can you please close the pull request?

PR #56, opened 2015-01-17
SLING-4327-added support for custom root paths for resource resolver

kciecierski:
https://issues.apache.org/jira/browse/SLING-4327

ovide ability to create a ResourceResolver which is aware of any mappings, for example by providing proper argument mappingPath. In case of AEM it would be possible to use publish instance mapping present under etc/publish.map to on author instance.
To achieve that the CommonResourceResolverFactoryImpl could be implementing methods getResourceResolver() and getAdministrativeResourceResolver() with additional argument defining the mapping location. The advantage of this solution is that the created ResourceResolver can be used many times with the same mappings. The drawback is that the mappings configuration will be found and cached when they resourceresolver will be used for the first time - there is no possibility to define the list of working mappings before.
The proposal of API extension:

ResourceResolver getResourceResolver(Map<String, Object> authenticationInfo, String customRootMap) throws LoginException;

ResourceResolver getAdministrativeResourceResolver(Map<String, Object> authenticationInfo, String customRootMap) throws LoginException;

### DIFF
--- a/bundles/api/src/main/java/org/apache/sling/api/resource/ResourceResolverFactory.java
+++ b/bundles/api/src/main/java/org/apache/sling/api/resource/ResourceResolverFactory.java
@@ -129,6 +129,27 @@ public interface ResourceResolverFactory {
      * returned will only have administrative privileges if the user identified
      * by the property has administrative privileges.
      * <p>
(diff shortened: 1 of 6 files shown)


rombert:

PR #66, opened 2015-03-09
SLING-4483 - Sightly: data-sly-resource does not properly resolve relative paths

vladbailescu:
Prepended current resource path to relative paths to make sure normalization doesn't break the paths. The same approach is used by <sling:include ../>


vladbailescu:
Applied by Radu Cotescu in [r1665219](https://svn.apache.org/viewvc?view=revision&revision=1665219) together with integration test.

PR #67, opened 2015-03-09
SLING-4484 - XSS POM references wrong scm URLs

vladbailescu:
Updated URLs to point to correct location.


vladbailescu:
Applied by Radu Cotescu in [r1665226](https://svn.apache.org/viewvc?view=revision&revision=1665226)

PR #72, opened 2015-03-13
SLING-4500: make sure that root has a primary type

santiagozky:
when using the mock resource resolver, the root node does not have a primary type. this causes issues on some cases, for example:
 Resource root= resourceResolver.getResource("/");
        assertFalse(root.isResourceType("app/resourceType"));

will throw a null pointer exception.
I've added jcr:root as the primary type of the root node and a test case for this.

### DIFF
--- a/testing/mocks/resourceresolver-mock/src/main/java/org/apache/sling/testing/resourceresolver/MockResourceResolverFactory.java
+++ b/testing/mocks/resourceresolver-mock/src/main/java/org/apache/sling/testing/resourceresolver/MockResourceResolverFactory.java
@@ -23,20 +23,18 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Stack;
 
-import org.apache.jackrabbit.JcrConstants;
 import org.apache.sling.api.resource.LoginException;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.apache.sling.api.resource.ResourceResolverFactory;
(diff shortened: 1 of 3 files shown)


testing/mocks/resourceresolver-mock/src/main/java/org/apache/sling/testing/resourceresolver/MockResourceResolverFactory.java
```diff
@@ -23,15 +23,20 @@
...
 import org.osgi.service.event.EventAdmin;
 
+import com.google.common.collect.ImmutableMap;
+
 /**
  * Simple resource resolver factory
  */
 public class MockResourceResolverFactory implements ResourceResolverFactory {
+    
+    private static final String JCR_ROOT="jcr:root";
```
rombert: Actually I think this should be [rep:root](http://jackrabbit.apache.org/api/2.1/org/apache/jackrabbit/spi/commons/name/NameConstants.html#REP_ROOT). jcr:root seems to be [a dummy name used in XML serialization](http://jackrabbit.apache.org/api/2.1/org/apache/jackrabbit/spi/commons/name/NameConstants.html#JCR_ROOT)

santiagozky: @rombert  thanks for noticing this. I checked the second link when I chose the name for it.

PR #74, opened 2015-03-19
SLING-4524: Provide the properties of the current Resource as a ValueMap...

olaf-otto:
https://issues.apache.org/jira/browse/SLING-4524: Provide the properties of the current Resource as a ValueMap via the sling:defineObjects tag


rombert:

olaf-otto:
https://issues.apache.org/jira/browse/SLING-4524 -> Won't fix

PR #78, opened 2015-03-25
SLING-4517 Update debian packaging to use crankstart.

bedge:
Patch switches debian packaging over to using crankstart and sling-s3 contrib projects.
Allows user to install a generic sling server, then edit one config file to select one of:
    tar files for nodes and data
    tar files for nodes, s3 for data
    mongo for nodes and data
    mongo for nodes, s3 for data


rombert:

PR #79, opened 2015-03-27
Log appropriate error message if RUNMODE is not set on initial startup.

bedge:
This package requires the user to choose the mode of operation. If it's not set it was just barfing without a useful message. This adds a more helpful message as to why it didn't start.


rombert:

PR #81, opened 2015-03-31
SLING-4557 - Add JSON and XML validation to the XSS Protection API

vladbailescu:
- Added validator methods to the API
- Added JSON validation implementation using Apache Commons JSON
- Added XML validation implementation using SAX
- Added unit tests

### DIFF
--- a/contrib/extensions/xss/src/main/java/org/apache/sling/xss/impl/XSSAPIImpl.java
+++ b/contrib/extensions/xss/src/main/java/org/apache/sling/xss/impl/XSSAPIImpl.java
@@ -16,24 +16,36 @@
  ******************************************************************************/
 package org.apache.sling.xss.impl;
 
+import java.io.StringReader;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


contrib/extensions/xss/src/main/java/org/apache/sling/xss/impl/XSSAPIImpl.java
```diff
@@ -46,6 +58,12 @@
 
     private static final Pattern PATTERN_AUTO_DIMENSION = Pattern.compile("['\"]?auto['\"]?");
 
+    private static SAXParserFactory factory = SAXParserFactory.newInstance();
```
rombert: Wouldn't this be better implemented as a field in activate which is then set to null on deactivate? This minimises the potential for pinning classes in memory once the bundle is updated

vladbailescu: Indeed, thanks for the suggestion!

PR #83, opened 2015-04-08
SLING-4639 - SlingAdaptToAttrProcessor to support Sling Models in Thymeleaf scripting engine

m-kay:
- Added new Thymeleaf processor which adapts object specified in data-sling-adaptabe to class specified in data-sling-adaptTo and makes object available on the template in the variable specified in data-sling-var
- Element itself will be removed after successful processing
- Example usage: 

``` html
<!--/*/ <div data-sling-adaptTo="org.apache.sling.api.resource.ValueMap" data-sling-var="properties" data-sling-adaptable="${resource}"></div> /*/-->
<h1 data-th-text="${properties['jcr:title']}">Title</h1>
```

---
##### SlingModels example

``` java
@Model(adaptables = {Resource.class})
public class MyModel {

    @Inject @Named("jcr:title")
    @Default(values = "Standard Titel")
    private String title;

    public String getTitle(){
        return this.title;
    }

}
```

``` html
<!--/*/ <div data-sling-adaptTo="com.thymeleaf.models.MyModel" data-sling-var="model" data-sling-adaptable="${resource}"></div> /*/-->
<h1 data-th-text="${model.title}">Title</h1>
```

### DIFF
--- a/contrib/scripting/thymeleaf/pom.xml
+++ b/contrib/scripting/thymeleaf/pom.xml
@@ -103,6 +103,12 @@
       <version>2.2.8</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.sling</groupId>
+      <artifactId>org.apache.sling.commons.classloader</artifactId>
+      <version>1.3.2</version>
+      <scope>provided</scope>
+    </dependency>
     <!-- Apache Felix -->
     <dependency>
       <groupId>org.apache.felix</groupId>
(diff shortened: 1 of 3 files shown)


m-kay:
Can be done with standard thymeleaf dialect:

``` html
<html data-th-with="page=${resource.adaptTo(@org.apache.sling.samples.fling.Page@class)}">
```

see: https://github.com/apache/sling/tree/trunk/samples/fling

PR #84, opened 2015-04-09
Why isn't MockSlingHttpServletRequest be adaptable to Resource.class?

georgkoester:
Does this make sense?

### DIFF
--- a/testing/mocks/sling-mock/src/test/java/org/apache/sling/testing/mock/sling/servlet/MockSlingHttpServletRequestTest.java
+++ b/testing/mocks/sling-mock/src/test/java/org/apache/sling/testing/mock/sling/servlet/MockSlingHttpServletRequestTest.java
@@ -18,29 +18,6 @@
  */
 package org.apache.sling.testing.mock.sling.servlet;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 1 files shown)


rombert:
It makes sense IMO. Since you already created the test, can you also add the implementation? Just go for the simplest possible fix and make sure to keep unrelated changes like moving imports around to a minimum.

Also, please file an issue at https://issues.apache.org/jira/browse/SLING so we can track the inclusion and your intention to contribute. Then the commit title should include that issue , e.g.

SLING-XYZ - Implement MockSlingHttpServletRequest.adaptTo(Resource.class)

georgkoester:
Hi Robert,

please review: https://github.com/apache/sling/pull/85

Cheers!
Georg

----- Ursprüngliche Mail -----
Von: "Robert Munteanu" notifications@github.com
An: "apache/sling" sling@noreply.github.com
CC: "Georg Koester" georg@deverado.de
Gesendet: Freitag, 10. April 2015 17:56:59
Betreff: Re: [sling] Why isn't MockSlingHttpServletRequest be adaptable to Resource.class? (#84)

It makes sense IMO. Since you already created the test, can you also add the implementation? Just go for the simplest possible fix and make sure to keep unrelated changes like moving imports around to a minimum.

Also, please file an issue at https://issues.apache.org/jira/browse/SLING so we can track the inclusion and your intention to contribute. Then the commit title should include that issue , e.g.

SLING-XYZ - Implement MockSlingHttpServletRequest.adaptTo(Resource.class)

---

PR #85, opened 2015-04-10
Sling-4607 - Implement MockSlingHttpServletRequest.adaptTo(Resource.clas...

georgkoester:
...s)


bundles/commons/testing/src/main/java/org/apache/sling/commons/testing/sling/MockSlingHttpServletRequest.java
```diff
@@ -398,6 +397,9 @@ public void setCharacterEncoding(String env) {
     }
 
     public <AdapterType> AdapterType adaptTo(Class<AdapterType> type) {
+        if (type == null && type.isAssignableFrom(Resource.class)) {
```
kwin: this doesn't look right. If type is null this line leads to an NPE!

georgkoester:
Thanks for checking @kwin , I added the test I was trying to find!

georgkoester:
Invalid, API doesn't work this way in Sling see SLING-4607.

PR #90, opened 2015-05-14
SLING-4715 - Event filter in ResourceEventDistributionTrigger is wrongly configured

maciusio:
You are setting event filter first to a given path and then to not handle events from other sling instances. But actually the path filter has been overriden. Probbaly you meant to have this filter configured like this.


tteofili:
pull request merged in r1679386

PR #106, opened 2015-10-15
SLING-5146 Add JSONObject/JSONArray support to JSONWriter

visiongeist:
Adds support for JSONObject and JSONArray to JSONWriter


rombert:
Thanks for the PR @visiongeist ! Some unit tests to go with this would be great. There are examples for this in the module but it those aren't clear enough let me know

PR #109, opened 2015-10-28
sling authenticator handle path that starts with the same prefix

santiagozky:
SLING-4701sling authenticator can handle paths that starts with the same prefix


rombert:
@santiagozky  - the Jira issue has been resolved, can you please close the pull request?

PR #110, opened 2015-11-09
SLING-5266  Export  org.mozilla.javascript.ast  package

welcomemandeep:
Currently Sling Scripting Javascript Support only exports org.mozilla.javascript package which is sufficient for evaluating javascript using Rhino but to convert expressions to  parsed AST we need org.mozilla.javascript.ast

### DIFF
--- a/bundles/scripting/javascript/pom.xml
+++ b/bundles/scripting/javascript/pom.xml
@@ -59,6 +59,8 @@
                         </Import-Package>
                         <Export-Package>
                             org.apache.sling.scripting.javascript,
+                            org.mozilla.javascript;version=1.1.0,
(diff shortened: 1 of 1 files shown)


bundles/scripting/javascript/pom.xml
```diff
@@ -59,6 +59,8 @@
                         </Import-Package>
                         <Export-Package>
                             org.apache.sling.scripting.javascript,
+                            org.mozilla.javascript;version=1.1.0,
+                            org.mozilla.javascript.ast;version=1.1.0,
                             org.mozilla.javascript;
```
nitin-nizhawan, line 64: @welcomemandeep it this somehow old code? line without version=1.1.0 should not be there

PR #115, opened 2015-11-26
SLING-5335: maven-sling-plugin support for SlingPostServlet usage.

kemuri-9:
- Allow usage of SlingPostServlet for install and uninstall operations.

### DIFF
--- a/tooling/maven/maven-sling-plugin/src/main/java/org/apache/sling/maven/bundlesupport/AbstractBundleInstallMojo.java
+++ b/tooling/maven/maven-sling-plugin/src/main/java/org/apache/sling/maven/bundlesupport/AbstractBundleInstallMojo.java
@@ -61,7 +61,7 @@ abstract class AbstractBundleInstallMojo extends AbstractBundlePostMojo {
     /** The fs resource provider factory. */
     private static final String FS_FACTORY = "org.apache.sling.fsprovider.internal.FsResourceProvider";
     /** Mime type for json response. */
-    private static final String JSON_MIME_TYPE = "application/json";
+    protected static final String JSON_MIME_TYPE = "application/json";
     /** Http header for content type. */
     private static final String HEADER_CONTENT_TYPE = "Content-Type";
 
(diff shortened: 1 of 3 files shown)


rombert:
Thanks for the submission, overall it looks good. A couple of notes:
- can you please add a sanity check which ensures that we don's select two conflicting options? `sling.useSlingPost` and `sling.usePut` can't both be true. And come to think of it, perhaps we should add a String parameter which describes the method used, e.g. `sling.deploy.method` which initially takes the values `webdav,webconsole,postServlet`
- would it make sense to extract the if/else statements into an interface with three implementations, as outlined above ( that would also make writing tests in the future much easier )

kemuri-9:
Sure a sanity check could be added to raise an error if both usePut and useSlingPost are both true. I was thinking about adding this, but wasn't sure if it was expected/appropriate.

As for adding a new deploy method, this would add some simplicity going forward for clarity purposes, but is it expected that this is an additional parameter and usePut still exists for backwards compatibility purposes so we would need to support both ways of setting, or would it be expected that usePut is removed, breaking backwards compatibility?

rombert:
> Sure a sanity check could be added to raise an error if both usePut and useSlingPost are both true. I was thinking about adding this, but wasn't sure if it was expected/appropriate.

I think it's better to fail hard. For instance if usePut is set to true in a parent pom and useSlingPost locally, it can get quite confusion.

> is it expected that this is an additional parameter and usePut still exists for backwards compatibility purposes

My suggestion would be to always obey the new parameter if non-null, and fall back to checking `usePut` if needed.

kemuri-9:
> I think it's better to fail hard. For instance if usePut is set to true in a parent pom and useSlingPost locally, it can get quite confusion.
> My suggestion would be to always obey the new parameter if non-null, and fall back to checking usePut if needed.

So instead of adding the `useSlingPost` boolean option, add the more flexible (future friendly) `sling.deploy.method` String option, which is allowed to be not specified.
So this seems to logically break down to 
- If `sling.deploy.method` is specified use that as specified (error out if invalid)
- if `sling.deploy.method` is not specified, check for `usePut` (backwards compatibility).
  - `usePut` is true -> use WebDAV
  - `usePut` is false -> use Felix Webconsole.

Or am I missing something here?

rombert:
Looks good to me

kemuri-9:
Added an additional commit here for switching from the `useSlingPost` boolean parameter to the `sling.deploy.method` String type parameter.
Let me know if the two commits should be squashed for convenience.

I agree that it would be likely better in the long term to switch things around to declare an interface for the deployment strategy and have different implementations, but given how invasive such a change would be, it warrants more discussion/collaborated thought on coming to a desired design strategy.

kemuri-9:
Actually question I have here now is, is the documentation at http://sling.apache.org/site/sling.html automatically generated from the Mojo, or is there somewhere that manual updates should be made for this change?

PR #120, opened 2015-12-22
SLING-5398: Fixing valid file check so it will also working on windows

bpauli:
On an windows environment the isValid method always return false because the root path of the fsclassloader may contain relative path information e.g. D:\aem\crx-quickstart\opt\helpers....\launchpad\felix\bundle445\data\classes


rombert:
@bpauli - thanks for the pull request. For tracking purposes, can you please also open a correspoding SLING jira?

bpauli:
done https://issues.apache.org/jira/browse/SLING-5398

PR #123, opened 2016-02-13
SLING-5517 Supports startsWith in Exclude paths in Content Disposition

welcomemandeep:


asanso:
sorry to close as wont't fix. We explicitly decided do not introduce this behavior.

rombert:

PR #130, opened 2016-03-17
Pull Request for Eclipse Independent Resource Change Command Factory

schaefa:
This pull request provides the New Resource Change Command Factory, its support class and some test classes copied from eclipse-test module. It also contains a fix for the RAT plugin to exclude IntelliJ module files (.iml).

### DIFF
--- a/bundles/extensions/models/impl/src/main/java/org/apache/sling/models/impl/AdapterImplementations.java
+++ b/bundles/extensions/models/impl/src/main/java/org/apache/sling/models/impl/AdapterImplementations.java
@@ -102,38 +102,52 @@ final class AdapterImplementations {
     }
     
     /** Add implementation mapping for the given model class (implementation is the model class itself).
-     * Only used for testing purposes. Use {@link #add(Class, Class)} in case you want to register a different implementation.
+     * Only used for testing purposes. Use {@link #addAll(Class, Class...)} in case you want to register a different implementation.
      * @param modelClasses the model classes to register
      */
     protected void addClassesAsAdapterAndImplementation(Class<?>... modelClasses) {
         for (Class<?> modelClass : modelClasses) {
-            add(modelClass, modelClass);
+            addAll(modelClass, modelClass);
         }
     }
     
     /**
-     * Add implementation mapping for the given adapter type.
-     * @param adapterType Adapter type
+     * Add implementation mapping for the given adapter types.
      * @param implType Implementation type
+     * @param adapterTypes Adapter types
+     * @result true if adapters were successfully added
      */
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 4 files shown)


rombert:
Thanks a lot for the pull request @schaefa !

Looking at the changes I see that these new classes are not integrated at all in the project. To make it possible to test and for me to understand how they fit in I would like to ask you to actually 'plug in' these classes where they are needed in the current project.

schaefa:
I cannot do that as I don't have the time to get up to speed, the knowledge and also not the access to the source code in question.

That said I did a pseudo implementation (based on my Eclipse wrappers) inside the AEM Tooling Plugin for IntelliJ (https://github.com/headwirecom/aem-ide-tooling-4-intellij/tree/develop/src/main/java/com/headwire/aem/tooling/intellij/io/eclipse). This should give you a hint on how I envisioned the implementation inside Eclipse. It is for sure not perfect as I cut corners in my wrappers.

rombert:
I see. So this is rather a proposal on how the APIs can evolve and a starting point for a discussion? And not used in the IntelliJ part either?

schaefa:
No, this isn't a proposal and it is used in IntelliJ. The parent folder of the one I mentioned contains the IntelliJ code. The eclipse folder is a proposal on how Eclipse could implement the API (SlingResource etc).

rombert:
I see. Well, to properly incorporate I would need to use them in the Eclipse tooling - it makes no sense to include them otherwise.

I'll start to do that, but don't expect them to be done overnight, and I might come back to this PR and comment about various things to further my understanding.

I might cherry-pick stuff and apply some minor changes if I think it fits the project better. How can I validate that I did not break the IntelliJ part?

schaefa:
You can't. If you leave the API intact it will work otherwise it for sure it won't. That said I should be able to make it work as it did shoehorning the eclipse code into IntelliJ in the first version.

I would suggest you have a look at my proposal of the eclipse implementation inside the IntelliJ project. I am pretty sure I messed a few things up (mostly because I couldn't find the source code) but it should give you an idea. We could do a screen share so we can discuss this if you have questions to expedite this (I am on PDT).

rombert:
I might be able to set this up on an IntelliJ environment ( have zero experience with that ) , but the point remains that we will need to iterate on the changes until we have a final version and make sure that everything still works for both Eclipse and IntelliJ.

The only way to do this is incrementally including your changes in the Eclipse part, and any changes that we agree on should be included in the IntelliJ part as well. But expect that to take some time...

schaefa:
You don’t need to setup an IntelliJ environment (will take some time as I am not sure what should be included into GIT). That said the changes in the code should not affect Eclipse as all changes are separate.

So we could merge my code into a separate branch and then integrate over it until we are both happy with it.

Cheers - Andy

PR #131, opened 2016-03-22
Documented the internals of Sling Event, with help from Carsten

ieb:
Documentation for Sling Events, will commit to svn if ok.


ieb:
@cziegeler  fyi, Documentation.

PR #132, opened 2016-03-24
SLING-5629: do not prepend servlet context path on the target

glucazeau:
Remove prepend of servlet context in redirectAfterLogout, as it is required to be in the redirection target by AuthUtil.isRedirectValid

### DIFF
--- a/bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
+++ b/bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
@@ -1414,7 +1414,7 @@ public class SlingAuthenticator implements Authenticator,
 
         // redirect to there
         try {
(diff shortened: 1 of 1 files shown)


bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
```diff
@@ -1414,7 +1414,7 @@ private void redirectAfterLogout(final HttpServletRequest request,
 
         // redirect to there
         try {
-            response.sendRedirect(request.getContextPath() + target);
+            response.sendRedirect(target);
```
kwin, line 1417: This at least is not correct in case the redirect is not valid (then target is "/" and this is obviously lacking the context path).

glucazeau:
Fixed by Carsten Ziegeler

PR #133, opened 2016-04-15
SLING-2781

mmanski:
Now valuemaps hold the same state by sharing a common cache. Unit test have been added to test the common state of ValueMap and JcrModifialbeValueMap, other unit tests are passing ofc


rombert:

PR #135, opened 2016-04-22
SLING-3543 solution proposal

mmanski:
This is a proposal of handling 3543. Console plugin makes real request for the selected path with a special selector, which is showing available scripting variables for the current path in json format. When we want to access scripting variables when for a path that is not supported by any of scripting engines, 400 is being thrown. 
example


bundles/scripting/core/src/main/resources/scriptingvariables/ui/scriptingvariables.js
```diff
@@ -0,0 +1,79 @@
...
+				{
+					type: 'GET'
+				}
+		).success(
+				function (data) {
+					$('#response').html(renderContent(data));
+				}
+		).fail(
+				function () {
+					$('#response').html('No scirpting context available under provided path.');
```
kwin: small typo

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/ScriptingContextValidationServlet.java
```diff
@@ -0,0 +1,52 @@
...
+        selectors = "bvpvars",
+        methods = "GET"
+)
+public class ScriptingContextValidationServlet extends SlingSafeMethodsServlet implements OptingServlet {
+
+    @Reference
+    private SlingScriptEngineManager scriptEngineManager;
+
+    @Override
+    protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse response) throws ServletException,
```
kwin, line 47: since this is not self-explanatory, please add some javadoc to the servlet and also to the doGet, which code is answering the requests

bundles/scripting/core/src/main/resources/scriptingvariables/ui/scriptingvariables.js
mmanski: fixed

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/ScriptingContextValidationServlet.java
mmanski, line 47: javadoc added

bundles/scripting/core/pom.xml
```diff
@@ -61,8 +61,9 @@
                 <configuration>
                     <instructions>
                         <Export-Package>
-                            org.apache.sling.scripting.core;version=2.1.0,
-                            org.apache.sling.scripting.core.servlet;version=1.0.0
+                            org.apache.sling.scripting.core;version=2.2.0,
+                            org.apache.sling.scripting.core.servlet;version=1.0.0,
+                            com.google.gson.*
```
kwin: Gson is not embedded in this bundle, how can you export it then?

bundles/scripting/core/pom.xml
```diff
@@ -192,6 +193,11 @@
             <version>2.4</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
kwin: Gson is not part of Sling yet. I would prefer to just rely on sling.commons.json here.

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/impl/DefaultSlingScript.java
```diff
@@ -371,21 +377,31 @@ public void setReader(Reader reader) {
...
-                    LOGGER.debug("Script {} was already cached.", scriptName);
+                    result = scriptEngine.eval(reader, ctx);
                 }
-                result = cachedScript.getCompiledScript().eval(ctx);
             } else {
-                result = scriptEngine.eval(reader, ctx);
+                acquireScriptingVariables(ctx, bvpVariables, 100, scriptEngine.getFactory().getEngineName());
+                acquireScriptingVariables(ctx, bvpVariables, 200, "GLOBAL");
+                acquireScriptingVariables(ctx, bvpVariables, -314, "SLING");
+                Gson  gson = new Gson();
```
kwin: Use commons.json to construct the json manually within the ScriptingVariableModel. We should be very careful before introducing additional 3rd party libraries.

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/impl/DefaultSlingScript.java
```diff
@@ -371,21 +377,31 @@ public void setReader(Reader reader) {
...
+                        LOGGER.debug("Script {} was already cached.", scriptName);
+                    }
+                    result = cachedScript.getCompiledScript().eval(ctx);
                 } else {
-                    LOGGER.debug("Script {} was already cached.", scriptName);
+                    result = scriptEngine.eval(reader, ctx);
                 }
-                result = cachedScript.getCompiledScript().eval(ctx);
             } else {
-                result = scriptEngine.eval(reader, ctx);
```
kwin, line 388: Some log debug would be helpful here

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/impl/DefaultSlingScript.java
```diff
@@ -371,21 +377,31 @@ public void setReader(Reader reader) {
...
-                CachedScript cachedScript = scriptCache.getScript(scriptName);
-                if (cachedScript == null) {
-                    ScriptNameAwareReader snReader = new ScriptNameAwareReader(reader, scriptName);
-                    CompiledScript compiledScript = ((Compilable) scriptEngine).compile(snReader);
-                    cachedScript = new CachedScriptImpl(scriptName, compiledScript);
-                    scriptCache.putScript(cachedScript);
-                    LOGGER.debug("Adding {} to the script cache.", scriptName);
+            Object result = null;
+            String selectorString = props.getRequest().getRequestPathInfo().getSelectorString();
+            if(StringUtils.isBlank(selectorString) || !selectorString.contains(BVP_VARIABLES_SELECTOR)) {
```
kwin, line 385: what if someone defines a dedicated script to answer requests with selector "bvpvars"? We should at least log a big warning in that case.

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/impl/DefaultSlingScript.java
```diff
@@ -96,6 +99,8 @@
     private static final Set<String> PROTECTED_KEYS =
         new HashSet<String>(Arrays.asList(REQUEST, RESPONSE, READER, SLING, RESOURCE, RESOLVER, OUT, LOG));
 
+    private static final String BVP_VARIABLES_SELECTOR = "bvpvars";
```
kwin, line 105: what does the p stand for in bvpvars?

mmanski, line 105: BVP - binding value provider

bundles/scripting/core/src/main/java/org/apache/sling/scripting/core/impl/DefaultSlingScript.java
```diff
@@ -371,21 +380,34 @@ public void setReader(Reader reader) {
...
                 } else {
-                    LOGGER.debug("Script {} was already cached.", scriptName);
+                    result = scriptEngine.eval(reader, ctx);
                 }
-                result = cachedScript.getCompiledScript().eval(ctx);
             } else {
-                result = scriptEngine.eval(reader, ctx);
+                LOGGER.warn("Found " + BVP_VARIABLES_SELECTOR + " selector in the request URI. "
+                        + "Script will not be evaluated. Returning available scripting variables as json...");
+                acquireScriptingVariables(ctx, bvpVariables, 100, scriptEngine.getFactory().getEngineName());
```
kwin, line 405: The according constants from ScriptContext (i.e. ENGINE_SCOPE, GLOBAL_SCOPE) and SlingScriptConstants.SLING_SCOPE should be used instead of the magic integers

PR #143, opened 2016-05-23
SLING-5741 discarding complete ResourceBundleCache in case of JSON di…

anagarwa:
…ctionary

### DIFF
--- a/bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
+++ b/bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
@@ -228,6 +228,27 @@ public class JcrResourceBundleProvider implements ResourceBundleProvider, EventH
                         break;
                     }
                 }
+
+                Resource resource = resourceResolver.getResource(path);
+                ValueMap valueMap = resource.adaptTo(ValueMap.class);
+
+                if (hasMixin(valueMap, JcrResourceBundle.MIXIN_MESSAGE)) {
+                    resource = resource.getParent();
+                    valueMap = resource.adaptTo(ValueMap.class);
+                }
+
+                if (hasMixin(valueMap, JcrResourceBundle.MIXIN_LANGUAGE) && valueMap.containsKey(JcrResourceBundle.PROP_LANGUAGE)) {
+                    String language = valueMap.get(JcrResourceBundle.PROP_LANGUAGE, String.class);
+                    Locale currLocale = toLocale(language);
+                    if (currLocale.equals(defaultLocale)) {
+                        scheduleReloadBundles(true);
+                        return;
+                    }
+                    discardBundlesWithLocalePrefix(language);
+                    return;
+                }
+
+
                 // may be a completely new dictionary
                 if (isDictionaryResource(path, event)) {
                     scheduleReloadBundles(true);
--- a/bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
+++ b/bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
@@ -197,9 +197,19 @@ public class JcrResourceBundleProvider implements ResourceBundleProvider, EventH
             // invalidate everything
             if (languageRootPaths.contains(path)) {
                 log.debug(
-                        "handleEvent: Detected change of cached language root '{}', removing all cached ResourceBundles",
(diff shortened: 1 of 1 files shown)


bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
```diff
@@ -197,9 +197,19 @@ public void handleEvent(final org.osgi.service.event.Event event) {
...
             if (languageRootPaths.contains(path)) {
                 log.debug(
-                        "handleEvent: Detected change of cached language root '{}', removing all cached ResourceBundles",
+                        "handleEvent: Detected change of cached language root '{}', removing all cached ResourceBundle that are affected",
                         path);
-                scheduleReloadBundles(true);
+
+                for (JcrResourceBundle bundle : resourceBundleCache.values()) {
+                    if (bundle.getLanguageRootPaths().contains(path)) {
+                        if (bundle.getLocale().equals(defaultLocale)) {
```
kwin: The whole if clause is IMHO not necessary, as your method discardBundleWithLocalePrefix will scheduleReloadBundles automatically in case the locale is the same.

anagarwa:
@kwin IMO "if" clause is needed to check which bundle does the path (for which event is generated) belongs to.

bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
```diff
@@ -226,6 +257,14 @@ public void handleEvent(final org.osgi.service.event.Event event) {
         }
     }
 
+    private void discardBundlesWithLocalePrefix(String localePrefix) {
```
kwin, line 260: This is for sure not enough. You need to consider the base name as well

bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
```diff
@@ -226,6 +257,14 @@ public void handleEvent(final org.osgi.service.event.Event event) {
...
+        for (JcrResourceBundle bundle : resourceBundleCache.values()) {
+            if (bundle.getLocale().toString().startsWith(localePrefix)) {
+                scheduleReloadBundle(bundle);
```
kwin, line 263: Actually scheduleReloadBundle already takes care of clearing dependent cached bundles

bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
```diff
@@ -197,9 +197,19 @@ public void handleEvent(final org.osgi.service.event.Event event) {
             // invalidate everything
             if (languageRootPaths.contains(path)) {
                 log.debug(
-                        "handleEvent: Detected change of cached language root '{}', removing all cached ResourceBundles",
+                        "handleEvent: Detected change of cached language root '{}', removing all cached ResourceBundle that are affected",
                         path);
-                scheduleReloadBundles(true);
+
+                for (JcrResourceBundle bundle : resourceBundleCache.values()) {
```
kwin, line 203: In case the locale of an existing dictionary was changed, the new dictionary would not correctly being preloaded here, because you call scheduleReloadBundle with the old locale only.

bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
anagarwa, line 260: @kwin IMO if there is a change in dictionary de_AT and we need to reload a bundle with locale de_AT all of its other variant like de_AT_MAC but we do not need to reload bundle with locale as de. 
More ever scheduling a bundle to reload will also reload parent bundle so I am not sure which case I am missing here. 
Can you please let me know on this ?

thanks,

kwin, line 260: I was referring to basename and you are only referring to locales. We are talking about different things here. basename is orthogonal to locales.

bundles/extensions/i18n/src/main/java/org/apache/sling/i18n/impl/JcrResourceBundleProvider.java
anagarwa, line 203: @kwin IMO that will not be as frequent case as updating a translation So in that case we can discard  the complete cache. Will that be okay ?

rombert:

PR #144, opened 2016-05-30
SLING-5752 - SimpleHttpDistributionTransport does not leverage configured scheme from distribution endpoint

tmaret:
- Pass scheme from distribution endpoint when building Executor's HttpHost


tteofili:
I've merged this in r1746247

PR #145, opened 2016-06-09
SLING-5768 Introduce rep:slingResourceTypes

ghenzler:
Introduce rep:slingResourceTypes as extension to Oak permission system


contrib/extensions/sling-oak-restrictions/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+				<artifactId>maven-bundle-plugin</artifactId>
+				<extensions>true</extensions>
+			</plugin>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>animal-sniffer-maven-plugin</artifactId>
+				<configuration>
+					<signature>
+						<groupId>org.codehaus.mojo.signature</groupId>
+						<artifactId>java16</artifactId>
```
rombert: This should not be needed if you inherit from the sling parent pom.

contrib/extensions/sling-oak-restrictions/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.jackrabbit</groupId>
+			<artifactId>oak-core</artifactId>
+			<version>1.2.7</version>
```
rombert, line 56: please make the scope provided, since you're importing transitive dependencies in the classpath, which makes it hard to see the actual dependencies.

You're using guava in the classes but they're not explicitly imported.

contrib/extensions/sling-oak-restrictions/src/main/java/org/apache/sling/slingoakrestrictions/impl/ResourceTypePattern.java
```diff
@@ -0,0 +1,139 @@
...
+        
+        this.resourceTypes = ImmutableSet.copyOf(plainResourceTypes);
+        this.resourceTypesByPath = ImmutableMap.copyOf(resourceTypesByPath);
+        LOG.debug("pattern setup with resourceTypes={}, resourceTypesByPath=", this.resourceTypes, this.resourceTypesByPath);
+    }
+
+    @Override
+    public boolean matches(@Nonnull Tree tree, @Nullable PropertyState property) {
+        String actualResourceType = TreeUtil.getString(tree, SLING_RESOURCE_TYPE);
+        boolean isResourceTypeMatch = resourceTypes.contains(actualResourceType);
```
rombert: Have you considered allowing for resource type inheritance?

contrib/extensions/sling-oak-restrictions/src/main/java/org/apache/sling/slingoakrestrictions/impl/SlingRestrictionProviderImpl.java
```diff
@@ -0,0 +1,107 @@
...
+                    return resourceTypePattern;
+                }
+            }
+        }
+
+        return RestrictionPattern.EMPTY;
+    }
+
+    @Override
+    public void validateRestrictions(String oakPath, @Nonnull Tree aceTree) throws AccessControlException {
```
rombert: is this needed?

contrib/extensions/sling-oak-restrictions/src/main/java/org/apache/sling/slingoakrestrictions/impl/ResourceTypePattern.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private final Set<String> resourceTypes;
+    private final Map<String,Set<String>> resourceTypesByPath;
+
+    ResourceTypePattern(@Nonnull Iterable<String> resourceTypesRaw) {
+        
+        List<String> plainResourceTypes = new ArrayList<String>();
+        Map<String,Set<String>> resourceTypesByPath = new HashMap<String,Set<String>>();
+        for (String resourceTypeRaw : resourceTypesRaw) {
+            if(resourceTypeRaw.contains("@")) {
```
rombert: Nitpick - if you're using Guava you might want to use `Splitter` as well.

contrib/extensions/sling-oak-restrictions/src/main/java/org/apache/sling/slingoakrestrictions/impl/SlingRestrictionProviderImpl.java
ghenzler: It's calling the base class at org.apache.jackrabbit.oak.spi.security.authorization.restriction.AbstractRestrictionProvider.validateRestrictions(String, Tree) which performs sanity checks - I think it makes sense.

rombert: My point was that you're overriding the method just to call the superclass method. Don't you achieve the same effect my simply removing the method?

ghenzler: You are right (the oak class was explicitly overriding because there was additional checks). Will remove it.

PR #146, opened 2016-06-13
SLING-5781 [i18n] generating locale from a string is incorrect

anagarwa:


kwin:
This one has been fixed differently. @anagarwa Can you close the PR?

PR #148, opened 2016-06-15
SLING-5788 - Implement a Sightly Maven Plugin

raducotescu:
- added a Sightly Maven Plugin based on the Sightly front-end compiler
  from [SLING-5787](https://github.com/apache/sling/pull/147)


raducotescu:
Patch applied in https://svn.apache.org/r1748819.

PR #150, opened 2016-06-28
SLING-5811 - Properly handle actual Resource's in Sightly data-sly-resource

vladbailescu:
- Added handling for actual Resource objects in data-sly-resource
- Added tests

### DIFF
--- a/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
+++ b/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
@@ -47,8 +47,6 @@ import org.apache.sling.scripting.sightly.compiler.RuntimeFunction;
 import org.apache.sling.scripting.sightly.extension.RuntimeExtension;
 import org.apache.sling.scripting.sightly.impl.utils.BindingsUtils;
 import org.apache.sling.scripting.sightly.render.RenderContext;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Runtime support for including resources in a Sightly script through {@code data-sly-resource}.
@@ -60,7 +58,6 @@ import org.slf4j.LoggerFactory;
 )
 public class ResourceRuntimeExtension implements RuntimeExtension {
 
-    private static final Logger LOG = LoggerFactory.getLogger(ResourceRuntimeExtension.class);
     private static final String OPTION_RESOURCE_TYPE = "resourceType";
     private static final String OPTION_PATH = "path";
     private static final String OPTION_PREPEND_PATH = "prependPath";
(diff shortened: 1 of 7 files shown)


bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
```diff
@@ -229,16 +238,24 @@ private Object getAndRemoveOption(Map<String, Object> options, String property)
 
     private void includeResource(final Bindings bindings, PrintWriter out, String script, String dispatcherOptions, String resourceType) {
         if (StringUtils.isEmpty(script)) {
-            LOG.error("Script path cannot be empty");
+            LOG.error("Resource path cannot be empty");
```
kwin: Can we rather throw an exception instead? That will also help with the new maven plugin.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
```diff
@@ -229,16 +238,24 @@ private Object getAndRemoveOption(Map<String, Object> options, String property)
...
             if (includeRes instanceof NonExistingResource || includeRes.isResourceType(Resource.RESOURCE_TYPE_NON_EXISTING)) {
                 includeRes = new SyntheticResource(request.getResourceResolver(), script, resourceType);
             }
+            includeResource(bindings, out, includeRes, dispatcherOptions, resourceType);
+        }
+    }
+
+    private void includeResource(final Bindings bindings, PrintWriter out, Resource includeRes, String dispatcherOptions, String resourceType) {
+        if (includeRes == null) {
+            LOG.error("Resource cannot be null");
```
kwin: I would prefer throwing an exception here as well. That way programming mistakes are detected earlier.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
vladbailescu: Makes sense and it's inline with https://github.com/apache/sling/blob/trunk/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/IncludeRuntimeExtension.java#L95

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
vladbailescu: Sure

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/ResourceRuntimeExtension.java
raducotescu: > That will also help with the new maven plugin.

@kwin: not really, as the plugin only receives the compilation warnings and errors. However I'd like to aim for consistency, so an exception makes more sense here.

PR #153, opened 2016-07-19
SLING-5792, SLING-5868, SLING-5795

anchela:
### Changes
- new interface 'AuthenticationRequirement'
- implemented by SlingAuthenticator
- additionally: annotations with helper classes in the impl package to improve readability
### Notes

with the implementation in SlingAuthenticator the values in the 'props' map change from array of AuthenticationRequirementHolder to a Set, which IMO corresponds better to the PathBasedHolderCache, which at some point has changed from ArrayList to SortedSet... but that for sure requires a careful review (also from a performance PoV)
### TODO
- the original code synchronized on 'props' before updating the PathBasedHolderCache, which itself was synchronized. That looks troublesome to me and with the patch this is separated.
- I would like to write some benchmarks to be able to compare performance of AuthenticationRequirement vs. the ServiceListener approach as well as the effect of moving from array to Set in the props-map. But: I couldn't find the right way of doing this in Apache Sling (see also SLING-5791 for a specific issue).
### Changes
- new module 'org.apache.sling.auth.requirement' in the contrib/auth space
- new public API 'LoginPathProvider'
- implementation based on Jackrabbit Oak that allows to mark subtrees with the set of configured supported paths to require authentication including an optional login path.
### TODO
- currently snapshot dependency to 'org.apache.sling.auth.core' due to usage of 'AuthenticationRequirement'
- repo-init that defines the service-user required for the default implementation
- Oak index definition that prevents traversal for initial query


rombert:

PR #156, opened 2016-07-20
SLING-5866 DefaultGetServlet obtains input stream for binary even if …

anagarwa:
…request is a HEAD


bundles/servlets/get/src/main/java/org/apache/sling/servlets/get/impl/helpers/HeadServletResponse.java
```diff
@@ -0,0 +1,127 @@
...
+        if (stream == null) {
+            stream = new NullServletOutputStream();
+        }
+
+        return stream;
+    }
+
+    @Override
+    public PrintWriter getWriter() {
+        if (stream != null) {
```
bdelacretaz, line 60: the null tests are inverted between getWriter and getOutputStream,

bundles/servlets/get/src/main/java/org/apache/sling/servlets/get/impl/helpers/HeadServletResponse.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private PrintWriter writer;
+
+    public HeadServletResponse(SlingHttpServletResponse wrappedResponse) {
+        super(wrappedResponse);
+    }
+
+    @Override
+    public ServletOutputStream getOutputStream() {
+        if (writer != null) {
```
bdelacretaz, line 47: the null tests are inverted between getWriter and getOutputStream,

anagarwa:
@bdelacretaz  as mentioned [here](http://docs.oracle.com/javaee/5/api/javax/servlet/ServletResponse.html#getOutputStream%28%29) "Either this method or getWriter() may be called to write the body, not both." So IMO, null test should be fine here . Moreover this is same as already done in sling [DefaultHeadServlet](https://github.com/apache/sling/blob/trunk/bundles/servlets/get/src/main/java/org/apache/sling/servlets/get/impl/DefaultHeadServlet.java#L133) . 
Please let me know if this is still a concern

rombert:

bdelacretaz:
It looks like this has been applied to our svn head and can be safely deleted.

PR #157, opened 2016-07-21
SLING-5874 improving timeout handling

ghenzler:


kwin:
This has been applied. @ghenzler Please close.

PR #158, opened 2016-07-22
SLING-5891 Provide compatibility for oak 1.2-1.5

ghenzler:
Sorry for the formatting noise  (now it is all spaces).

### DIFF
--- a/contrib/extensions/sling-oak-restrictions/pom.xml
+++ b/contrib/extensions/sling-oak-restrictions/pom.xml
@@ -1,100 +1,114 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
-<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor 
-	license agreements. See the NOTICE file distributed with this work for additional 
-	information regarding copyright ownership. The ASF licenses this file to 
-	you under the Apache License, Version 2.0 (the "License"); you may not use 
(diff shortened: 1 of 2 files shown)


contrib/extensions/sling-oak-restrictions/pom.xml
```diff
@@ -1,100 +1,114 @@
...
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-scr-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
```
ghenzler, line 49: This is the part that has actually changed

contrib/extensions/sling-oak-restrictions/src/main/java/org/apache/sling/slingoakrestrictions/impl/SlingRestrictionProviderImpl.java
```diff
@@ -23,7 +23,6 @@
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import javax.jcr.security.AccessControlException;
```
ghenzler, line 26: Removed unused import

rombert:

PR #160, opened 2016-08-05
Bugfix/sling 5947

buuhuu:
I implementing the methods throwing UnsupportedOperationException when the MockSlingHttpServletRequest is used with the SlingRequestProcessor. 

Unfortunately there are some older commits included in this PR, which actually don't change anything. Hope thats not a problem.


bundles/extensions/servlet-helpers/src/main/java/org/apache/sling/servlethelpers/MockSlingHttpServletRequest.java
```diff
@@ -632,56 +634,111 @@ public void setRemotePort(int remotePort) {
...
     @Override
-    public Enumeration<String> getResponseContentTypes() {
-        throw new UnsupportedOperationException();
+    public StringBuffer getRequestURL() {
+        StringBuffer requestUrl = new StringBuffer();
+
+        requestUrl.append(this.getScheme());
+        requestUrl.append("://");
+        requestUrl.append(getServerName());
+        requestUrl.append(':');
```
kwin: the port should only be included if it is not 80 for scheme=http or 443 for scheme=https

bundles/extensions/servlet-helpers/src/test/java/org/apache/sling/servlethelpers/MockSlingHttpServletRequestTest.java
```diff
@@ -88,6 +88,44 @@ public void testSession() {
     }
```
kwin, line 89: A test for getRequestURL would be useful as well showing the default port behaviour

buuhuu:
Resolved.

PR #162, opened 2016-08-18
SLING-5975 - POST fails with modern multi-value parameter names containing brackets

roele:


rombert:

PR #163, opened 2016-09-01
SLING-5768 using sling:resourceTypesWithDescendants instead of sling:resourceTypesWithChildren now

ghenzler:


rombert:
@ghenzler  - can you please open a new SLING issue and add the reason for the rename there? If we keep adding stuff on top of SLING-5768 I fear we're going to lose track of the reasons for which we did the rename.

PR #169, opened 2016-09-07
SLING-6045 Moving module to contrib/jcr

ghenzler:


rombert:
Should this be closed now that we applied #173 ?

PR #172, opened 2016-09-13
SLING-6053 SlingAuthenticator identifies wrong sibling node with AuthenticationInfo

mkbrv:
https://issues.apache.org/jira/browse/SLING-6053

### DIFF
--- a/bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
+++ b/bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
@@ -543,9 +543,9 @@ public class SlingAuthenticator implements Authenticator,
             final Collection<AbstractAuthenticationHandlerHolder> holderList = holdersArray[m];
             if ( holderList != null ) {
                 for (AbstractAuthenticationHandlerHolder holder : holderList) {
-                    if (path.startsWith(holder.path)) {
+                    if (new AuthenticationHandlerPath(path, holder.path).isNodeRequiresAuthHandler()) {
                         log.debug("login: requesting authentication using handler: {}",
-                            holder);
+                                holder);
 
                         try {
                             done = holder.requestCredentials(request, response);
@@ -604,7 +604,7 @@ public class SlingAuthenticator implements Authenticator,
             final Collection<AbstractAuthenticationHandlerHolder> holderSet = holdersArray[m];
             if (holderSet != null) {
                 for (AbstractAuthenticationHandlerHolder holder : holderSet) {
-                    if (path.startsWith(holder.path)) {
+                    if (new AuthenticationHandlerPath(path, holder.path).isNodeRequiresAuthHandler()) {
                         log.debug("logout: dropping authentication using handler: {}",
                             holder);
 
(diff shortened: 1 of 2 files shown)


bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
```diff
@@ -1639,4 +1639,35 @@ protected AbstractAuthenticationHandlerHolder createHolder(String path,
...
     }
 
+    /**
+     * SLING-6053
+     * Auth handler is required for the node and for all his children.
+     * Solved the case when sibling node name startsWith same name.
+     * <p>
+     * Eg: node: /page & node /page1
+     */
+    static class AuthenticationHandlerPath {
```
kwin, line 1649: Is a simple private method just taking two strings in the SlingAuthenticator not enough?

bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
```diff
@@ -1639,4 +1639,35 @@ protected AbstractAuthenticationHandlerHolder createHolder(String path,
...
+        }
+
+        boolean isNodeRequiresAuthHandler() {
+            if (requestPath.equalsIgnoreCase(handlerPath)) {
+                return true;
+            }
+            if (requestPath.startsWith(handlerPath)) {
+                final String suffix = requestPath.substring(handlerPath.length());
+                //if is child node auth handler is applied;
+                return suffix.contains("/");
```
kwin, line 1670: what about we have `requestPath` = `page1\somesubepage` and `handlerPath`= `page`?  If I understand correctly, this would return `true`although it is supposed to return `false`.

mkbrv: I have added a test for this case.

kwin: Sorry for he confusion, of course I meant `page1/somesubpage` which would still return `true` although it is meant to return `false`.

bundles/auth/core/src/main/java/org/apache/sling/auth/core/impl/SlingAuthenticator.java
```diff
@@ -1657,11 +1657,15 @@ protected AbstractAuthenticationHandlerHolder createHolder(String path,
         }
 
         boolean isNodeRequiresAuthHandler() {
-            if (requestPath.equalsIgnoreCase(handlerPath)) {
+            final String requestPathNoSelectors = requestPath.contains(".") ?
```
kwin, line 1660: This condition is not correct. Just that a path contains a dot does not mean that the part before the first dot is the resource path. Basically for the path `/some/path.extension/test/some.other.html`, you cannot tell without knowing the repository structure whether there is a resource at `/some/path.extension/test/some`, which means that there is no suffix and the selector `other` and the extension `html` or in case there is a resource only at `/some/path` the part `extension` would be the extension, there would be no selector and `/test/some.other.html` would be the suffix. See also https://sling.apache.org/documentation/the-sling-engine/url-decomposition.html.

mkbrv: makes sense. An alternative would be to split the request path into an array by each level in the repository (split by "/"). Also split the handler path in a similar way and see if the last 2 nodes on each are on the same level and not equal.

mkbrv:
Added with commit:  2bd37e44f4bbdb0acab8354b960760b7ef3f97a7 

https://github.com/apache/sling/commit/2bd37e44f4bbdb0acab8354b960760b7ef3f97a7

PR #178, opened 2016-10-07
SLING-6110: use the proper bundleContext in the ServerSideTeleporter …

marcel-j:
…to retrieve the correct bundle headers


rombert:

PR #180, opened 2016-10-10
Ability to specify TTL for separate health check

drfits:
Added ability to specify TTL for separate health check in case of expensive computations.
If property "hc.ttl" specified within HC - it will be used as TTL for result, otherwise default TTL value

**Ex.**: in my case "hc" validating repository about 3-5 minutes therefore I couldn't specify TTL globally to not impact on other "hc" results. This check I couldn't execute by scheduler due to CPU loading, also results for this check results remain relevant for an hour.

Ticket for this request: https://issues.apache.org/jira/browse/SLING-6126

### DIFF
--- a/bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
+++ b/bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
@@ -68,10 +68,10 @@ public interface HealthCheck {
     String ASYNC_CRON_EXPRESSION = "hc.async.cronExpression";
 
     /**
-     * Optional service property: Time To Leave for health check {@link Result}.
+     * Optional service property: Time To Live for health check {@link Result}.
      * The value of this property must be of type {@link Long}.
      */
-    String TTL = "hc.ttl";
+    String RESULT_CACHE_TTL_IN_MS = "hc.resultCacheTtlInMs";
 
     /**
      * Execute this health check and return a {@link Result}
--- a/bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
+++ b/bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
@@ -66,7 +66,13 @@ public interface HealthCheck {
      * will be executed asynchronously using the cron expression provided.
      */
     String ASYNC_CRON_EXPRESSION = "hc.async.cronExpression";
-    
+
+    /**
+     * Optional service property: Time To Leave for health check {@link Result}.
+     * The value of this property must be of type {@link Long}.
+     */
+    String TTL = "hc.ttl";
+
     /**
      * Execute this health check and return a {@link Result}
      * This is meant to execute quickly, access to external
--- a/bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
+++ b/bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
@@ -47,6 +47,10 @@ public class HealthCheckMetadata {
 
     private final transient ServiceReference serviceReference;
 
+    private final long ttl;
+
+    private final static long NOT_CONFIGURED_TTL = -1;
+
     public HealthCheckMetadata(final ServiceReference ref) {
         this.serviceId = (Long) ref.getProperty(Constants.SERVICE_ID);
         this.name = (String) ref.getProperty(HealthCheck.NAME);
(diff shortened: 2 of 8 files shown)


bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
```diff
@@ -66,7 +66,13 @@
      * will be executed asynchronously using the cron expression provided.
      */
     String ASYNC_CRON_EXPRESSION = "hc.async.cronExpression";
-    
+
+    /**
+     * Optional service property: Time To Leave for health check {@link Result}.
+     * The value of this property must be of type {@link Long}.
+     */
+    String TTL = "hc.ttl";
```
ghenzler: Can the property be named `hc.resultCacheTtlInMs`? (would be in line with global config and clearly describes it's about the result cache, not about the HC execution time)

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
```diff
@@ -66,7 +66,13 @@
      * will be executed asynchronously using the cron expression provided.
      */
     String ASYNC_CRON_EXPRESSION = "hc.async.cronExpression";
-    
+
+    /**
+     * Optional service property: Time To Leave for health check {@link Result}.
```
ghenzler: TTL = Time to live (not leave)

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
```diff
@@ -47,13 +47,19 @@
 
     private final transient ServiceReference serviceReference;
 
+    private final long ttl;
```
ghenzler: We could use `Long resultCacheTtlInMs` - then the value `null` can indicate that it is not configured

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
```diff
@@ -47,13 +47,19 @@
...
+    private final static long NOT_CONFIGURED_TTL = -1;
+
     public HealthCheckMetadata(final ServiceReference ref) {
         this.serviceId = (Long) ref.getProperty(Constants.SERVICE_ID);
         this.name = (String) ref.getProperty(HealthCheck.NAME);
         this.mbeanName = (String) ref.getProperty(HealthCheck.MBEAN_NAME);
         this.title = getHealthCheckTitle(ref);
         this.tags = arrayPropertyToListOfStr(ref.getProperty(HealthCheck.TAGS));
         this.asyncCronExpression = (String) ref.getProperty(HealthCheck.ASYNC_CRON_EXPRESSION);
+        final Long ttlTemp = PropertiesUtil.toLong(ref.getProperty(HealthCheck.TTL), NOT_CONFIGURED_TTL);
```
ghenzler: see comment above

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
```diff
@@ -116,6 +122,14 @@ public ServiceReference getServiceReference() {
         return this.serviceReference;
     }
 
+    /**
+     * Get service {@link org.apache.sling.hc.api.Result} Time To Leave
+     * @return actual time in ms for {@link org.apache.sling.hc.api.Result} for leave in cache
+     */
+    public long getTtl() {
```
ghenzler: Property name `resultCacheTtlInMs` would be better

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
ghenzler: This property should also be made available via annotation https://github.com/drfits/sling/blob/trunk/bundles/extensions/healthcheck/annotations/src/main/java/org/apache/sling/hc/annotations/SlingHealthCheck.java

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
drfits: changed

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
drfits: changed

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
drfits: changed

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
drfits: changed

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/util/HealthCheckMetadata.java
drfits: changed

bundles/extensions/healthcheck/core/src/main/java/org/apache/sling/hc/api/HealthCheck.java
drfits: added

rombert:

drfits:
Already merged

PR #181, opened 2016-10-11
Issue/sling 5135

bdelacretaz:
Ready to commit this


bdelacretaz:
I have committed this separately as svn revision 1764281

PR #182, opened 2016-10-11
Adding dateFormat

heervisscher:
PR to test the flow...

### DIFF
--- a/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
+++ b/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
@@ -24,6 +24,7 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
 
+import org.apache.sling.scripting.sightly.SightlyException;
 import org.apache.sling.scripting.sightly.extension.RuntimeExtension;
 import org.apache.sling.scripting.sightly.render.RenderContext;
 import org.slf4j.Logger;
@@ -80,9 +81,7 @@ public class DateFormatExtension implements RuntimeExtension {
 			SimpleDateFormat formatter = new SimpleDateFormat(format);
 			return formatter.format(date);
 		} catch (Exception e) {
-			LOG.error("Error during formatting, date {0} with format {1]", date, format);
-			LOG.error("Error during formatting", e);
+			throw new SightlyException("Error during formatting of date", e);
 		}
-		return null;
 	}
 }
--- a/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
+++ b/bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
@@ -41,32 +41,44 @@ public class DateFormatExtension implements RuntimeExtension {
 		Object dateValue = arguments[0];
 
 		String dateFormat = String.valueOf(arguments[1]);
-
+		
+		String returnValue = null;
+		
 		if (dateValue instanceof GregorianCalendar) {
 
 			GregorianCalendar gregorianCal = (GregorianCalendar) dateValue;
-			SimpleDateFormat formatter = new SimpleDateFormat(dateFormat);
-			return formatter.format(gregorianCal.getTime());
+			returnValue = format(gregorianCal.getTime(), dateFormat);
 
 		} else if (dateValue instanceof Calendar) {
 
 			Calendar cal = (Calendar) dateValue;
-			String value = format(cal, dateFormat);
-			return value;
+			returnValue = format(cal.getTime(), dateFormat);
 
 		} else if (dateValue instanceof Date) {
 
-			SimpleDateFormat formatter = new SimpleDateFormat(dateFormat);
 			Date dateAsDate = (Date) dateValue;
-			return formatter.format(dateAsDate.getTime());
+			returnValue = format(dateAsDate, dateFormat);
 
-		} else {
-			return dateValue;
 		}
+		if ( returnValue == null ) {
+			LOG.trace("Returnvalue is null, returning original value");
+			return dateValue;
+		} else {
+			return returnValue;
+		}
+		
 	}
 
-	private String format(Calendar cal, String format) {
-		SimpleDateFormat formatter = new SimpleDateFormat(format);
-		return formatter.format(cal);
+	private String format(Date date, String format) {
+		LOG.trace("Formatting date {0}, with format {1}", date, format);
+
+		try {
+			SimpleDateFormat formatter = new SimpleDateFormat(format);
+			return formatter.format(date);
+		} catch (Exception e) {
+			LOG.error("Error during formatting, date {0} with format {1]", date, format);
+			LOG.error("Error during formatting", e);
+		}
+		return null;
 	}
 }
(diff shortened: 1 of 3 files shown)


bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	private String format(Date date, String format) {
+		LOG.trace("Formatting date {0}, with format {1}", date, format);
+
+		try {
+			SimpleDateFormat formatter = new SimpleDateFormat(format);
+			return formatter.format(date);
+		} catch (Exception e) {
+			LOG.error("Error during formatting, date {0} with format {1]", date, format);
```
kwin: I would prefer to throw a meaningful exception instead of just doing some logging here. Compare with https://issues.apache.org/jira/browse/SLING-4175.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
```diff
@@ -0,0 +1,88 @@
...
+
+			Calendar cal = (Calendar) dateValue;
+			returnValue = format(cal.getTime(), dateFormat);
+
+		} else if (dateValue instanceof Date) {
+
+			Date dateAsDate = (Date) dateValue;
+			returnValue = format(dateAsDate, dateFormat);
+
+		}
```
kwin: Please throw an exception in case `dateValue` is of none of the three supported types.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
heervisscher: thanks for the feedback, now throwing an exception

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
heervisscher: done

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	private String format(Date date, String format) {
+		LOG.trace("Formatting date {0}, with format {1}", date, format);
+
+		try {
+			SimpleDateFormat formatter = new SimpleDateFormat(format);
+			return formatter.format(date);
+		} catch (Exception e) {
+			throw new SightlyException("Error during formatting of date", e);
```
kwin: Could you give out the format string and the date in the exception?

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
```diff
@@ -0,0 +1,89 @@
...
+			Calendar cal = (Calendar) dateValue;
+			returnValue = format(cal.getTime(), dateFormat);
+
+		} else if (dateValue instanceof Date) {
+
+			Date dateAsDate = (Date) dateValue;
+			returnValue = format(dateAsDate, dateFormat);
+
+		} else {
+			throw new SightlyException("Input value is not of a date type");
```
kwin: Please give out the supported types explicitly.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
```diff
@@ -0,0 +1,89 @@
...
+	@Override
+	public Object call(final RenderContext renderContext, Object... arguments) {
+		ExtensionUtils.checkArgumentCount("dateFormat", arguments, 2);
+		Object dateValue = arguments[0];
+
+		String dateFormat = String.valueOf(arguments[1]);
+		
+		String returnValue = null;
+		
+		if (dateValue instanceof GregorianCalendar) {
```
kwin: A `GregorianCalendar` is inheriting from `Calendar`. Therefore this if block can be completely removed.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
```diff
@@ -0,0 +1,89 @@
...
+
+		} else if (dateValue instanceof Date) {
+
+			Date dateAsDate = (Date) dateValue;
+			returnValue = format(dateAsDate, dateFormat);
+
+		} else {
+			throw new SightlyException("Input value is not of a date type");
+		}
+		if ( returnValue == null ) {
```
kwin: I don't think this can now ever happen (due to the exception). Please remove the block.

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
heervisscher: done

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
heervisscher: done

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
heervisscher: done

bundles/scripting/sightly/engine/src/main/java/org/apache/sling/scripting/sightly/impl/engine/extension/DateFormatExtension.java
heervisscher: done:-)

vladbailescu:
Just an idea: how about defining a more generic `FormattingFilter`?
It could:
- either dispatch to all sorts of `FormattingExtension` based on `format*` option
- or just use the [Formatter](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html)? Like `${date @ formatter='...'}`. Unfortunately, the `format` option is already in use

heervisscher:
hi @vladbailescu ,

thanks for your feedback, given that 'format' is already used, I think it was more complex / confusing how this can be used.

kwin:
What is missing in this PR is 
1. Support for new Java8 date classes: http://www.oracle.com/technetwork/articles/java/jf14-date-time-2125367.html
2. The locale is not specified in the SimpleDateFormatter constructor. This should be taken from the request by default (and passed to the SimpleDateFormatter, https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html#SimpleDateFormat(java.lang.String,%20java.util.Locale). Also there should be a possibility to overwrite that with an additional option, similar to what we have for `i18n`, https://github.com/Adobe-Marketing-Cloud/htl-spec/blob/master/SPECIFICATION.md#123-i18n.

wolfpl:
What is also missing here is the timezone to express the date (it would default to the system default, which is not necessarily the time you want).

heervisscher:
hi @wolfpl , great suggestion! I added the support for timezones.

vladbailescu:
@raducotescu @heervisscher, this has been fixed at https://issues.apache.org/jira/browse/SLING-6102. We should probably close this.

rombert:

PR #185, opened 2016-11-09
SLING-5135 - Whitelist legit usages of loginAdministrative and administrative ResourceResolver

jsedding:
This pull request includes the following refactorings:

1. Replace `AbstractSlingRepositoryManager#getLoginAdminWhitelist()` with `AbstractSlingRepositoryManager#allowLoginAdministrativeForBundle(Bundle)`. This allows implementations of `AbstractSlingRepositoryManager` to implement this method as they see fit. This decouples the `LoginAdminWhitelist` service fro `AbstractSlingRepositoryManager`, in fact only `OakSlingRepositoryManager` needs this dependency, as it uses it in its implementation.

2. Refactor `JcrProviderStateFactory` to use an `AbstractSlingRepository` instance injected with the "usingBundle" (by means of a `ServiceFactory`) not only for `loginService` but also for `loginAdministrative`. This allows removing any reference to `LoginAdminWhitelist` and thus also the dependency from `o.a.s.jcr.resource` to `o.a.s.jcr.base`.

3. Rename `LoginAdminWhitelist` configuration properties as suggested by Oliver in SLING-5135


jsedding:
Applied changes.

PR #191, opened 2016-12-12
Added implementation for date and number formatting to HTL

vladbailescu:
* SLING-6102 - Adding formatter to HTL/Sightly dates  …
* SLING-6140 - Adding number formatting to Sightly/HTL

### DIFF
--- a/bundles/scripting/sightly/java-compiler/src/main/java/org/apache/sling/scripting/sightly/render/AbstractRuntimeObjectModel.java
+++ b/bundles/scripting/sightly/java-compiler/src/main/java/org/apache/sling/scripting/sightly/render/AbstractRuntimeObjectModel.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -68,7 +69,7 @@ public abstract class AbstractRuntimeObjectModel implements RuntimeObjectModel {
 
     @Override
     public boolean isDate(Object target) {
-        return (target instanceof Date);
+        return (target instanceof Date || target instanceof Calendar);
     }
 
     @Override
@@ -109,6 +110,8 @@ public abstract class AbstractRuntimeObjectModel implements RuntimeObjectModel {
     public Date toDate(Object object) {
         if (object instanceof Date) {
             return (Date)object;
+        } else if (object instanceof Calendar) {
+            return ((Calendar)object).getTime();
         }
         return new Date(0);
     }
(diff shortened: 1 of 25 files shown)


bundles/scripting/sightly/java-compiler/src/main/java/org/apache/sling/scripting/sightly/render/AbstractRuntimeObjectModel.java
```diff
@@ -67,6 +67,16 @@ public boolean isPrimitive(Object obj) {
     }
 
     @Override
+    public boolean isDate(Object target) {
+        return (target instanceof Date);
```
heervisscher: what about Calendar?

bundles/scripting/sightly/testing-content/src/main/resources/SLING-INF/apps/sightlyperf/test/sly-js-sync.html
```diff
@@ -18,6 +18,7 @@
     ${test.tag != null ? test.startTag : '' @ context = "unsafe"}
     ${test.text @ context = "text"}
     ${test.tag != null ? test.endTag : '' @ context = "unsafe"}
+    ${'yyyy-MM-dd HH:mm:ss.SSS' @ format = test.date}
```
heervisscher, line 21: I miss tests for date formatting with timezone with/without locale.
Also I don't see tests for number-formatting.

vladbailescu: The functionality tests are part of the HTL TCK (https://github.com/Adobe-Marketing-Cloud/htl-tck/pull/20/files#diff-05ef2ebd0db1e547f6e90524ab379328R39). The tests here are just for performance evaluation and I think this covers the worst scenario.

bundles/scripting/sightly/java-compiler/src/main/java/org/apache/sling/scripting/sightly/render/AbstractRuntimeObjectModel.java
vladbailescu: Since the HTL specification will include explicit option for timezone (and Calendar is a Date + Timezone) I don't think it's really that helpful to provide shorthands for Calendar object formatting (their internal timezone might clash with the user-specified one). The HTL syntax makes it trivial to derive date and timezone from a Calendar: `${'yyyy-MM-dd HH:mm' @ format=myCal.time, timezone=myCal.timeZone}`

heervisscher: As a developer you don't want to worry about this, you have a 'date' object. Whether this is a Date or Calendar.
Would be great if this can be implemented.
You can have a look at my implementation, this supports date as well as calendar.

vladbailescu: @heervisscher How about timezones included in Calendar? Do you suggest to use the value to populate the timezone option (if no explicit value is provided) or to ignore it completely?

heervisscher: @vladbailescu  Use it indeed as a default if no timezone is specified.

PR #195, opened 2017-01-05

nitin-nizhawan:
- Add support for restriction clause in parser
   - example acl with restriction spec
     - set ACL on /libs
         deny jcr:modifyProperties for user2 restriction(rep:itemNames,prop1,prop2)
       end
     - set ACL for user1,u2
         allow jcr:addChildNodes on /apps,/content restriction(rep:glob,/cat/*,*/cat,*cat/*)
       end


rombert:

PR #197, opened 2017-01-09
SLING-6428 - DateFormat is printed when date-value is empty

vladbailescu:
* Updated TCK to include tests for empty/undefined formatting input values
* Updated implementation to return empty when formatting input is null/undefined
* Added missing javadocs


vladbailescu:
@raducotescu, can you please take a look?

raducotescu:
@vladbailescu, I applied your patch but forgot to mark this PR as closed from the commit message. Could you please close the PR?

vladbailescu:
Thanks!

PR #198, opened 2017-01-25
SLING-6482: Fix Exception in Commons Log WebConsole

bjoernweide:
SLING-6482: Fix StringIndexOutOfBoundsException in Commons Log WebConsole by adding a length check for the path. 
https://issues.apache.org/jira/browse/SLING-6482


rombert:

PR #199, opened 2017-02-15
SLING-5457 - OsgiInstaller should retry to start bundles on failures

DominikSuess:
- adding 5 retry cycles in BundleUpdateTask for Exceptions

### DIFF
--- a/installer/core/src/main/java/org/apache/sling/installer/core/impl/tasks/BundleUpdateTask.java
+++ b/installer/core/src/main/java/org/apache/sling/installer/core/impl/tasks/BundleUpdateTask.java
@@ -36,6 +36,11 @@ public class BundleUpdateTask extends AbstractBundleTask {
 
     private static final String BUNDLE_UPDATE_ORDER = "50-";
 
+    private static final int MAX_RETRIES = 5;
+    
+    // keep track of retry attempts via temporary attribute stored in taskresource
+    private String ATTR_UPDATE_RETRY = "org.apache.sling.installer.core.impl.tasks.BundleUpdateTask.retrycount";
+
     public BundleUpdateTask(final TaskResourceGroup r,
                             final TaskSupport creator) {
         super(r, creator);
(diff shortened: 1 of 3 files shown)


kwin:
I think adding the retry handling to the OSGiInstallerImpl (and therefore let all tasks profit from it) makes a lot more sense. See my proposal in http://www.mail-archive.com/dev@sling.apache.org/msg60947.html.

DominikSuess:
@kwin   a change to osgiInstallerImpl would be a lot more intrusive and in fact the others wouldn't benefit at all because they already endlessly retry. The issue with the exception handling in BundleUpgradeTask was that the status for the ResourceTask was set to ignore. Instead of switching for endless retry mode (like BundleInstallTask does) I "delay" for a certain amount of retries. 

Getting that to a higher level would require a significant change as statehandling is currently done within the ResourceTasks. I suggest to go for this low impact fix now and consider consolidating the Installer mechanisms and revisiting the state handling over all midterm (as  the current way this is done is quite brittle).

kwin:
Ok, understood and agreed, but there is currently no call to retryHandler.scheduleRetry (like in https://github.com/apache/sling/blob/trunk/installer/factories/packages/src/main/java/org/apache/sling/installer/factory/packages/impl/PackageTransformer.java#L252). Therefore the retry is only done with the next cycle of the OSGi Installer (e.g. when a new bundle is supposed to be installed/updated). That seems to be wrong here, because it may cause a very long delay until the bundle update is actually retried. I would therefore go for a sleep and retry approach here.

karlpauls:
I'm going to apply this patch and do a release. 

@kwin I did comment a bit more on your points on SLING-5457.

DominikSuess:
@kwin the osgi installer cycle at the end compacts the persistentList in case any Resource is left the next cycle is immediately triggered (return ACTION.CYCLE;) This is why setting to IGNORE breaks the cycle but exiting a TaskResource unhandled causes the next cycle. Just check my IT which covers positive and negative cases and proves that multiple retries happen without any further intervention.

kwin:
Not sure I can follow here. I see that https://github.com/apache/sling/blob/533d13c2138cce10baa146f11199b1d8d3f99d5d/installer/core/src/main/java/org/apache/sling/installer/core/impl/EntityResourceList.java#L434 only returns `true` in case some resource in the resource group has been uninstalled. Therefore https://github.com/apache/sling/blob/5473afd0196b58d0e0395cf6a98296af0dcc664f/installer/core/src/main/java/org/apache/sling/installer/core/impl/OsgiInstallerImpl.java#L913 should return `false` in your case as well. Not sure why in your IT the according TaskFactory is indeed called multiple times.

If 

> the osgi installer cycle at the end compacts the persistentList in case any Resource is left the next cycle is immediately triggered (return ACTION.CYCLE;) 

would be true, we would have a loop in https://github.com/apache/sling/blob/8317c5d255192113587a1369571fe9ad2b7d8014/installer/factories/packages/src/main/java/org/apache/sling/installer/factory/packages/impl/PackageTransformer.java#L234.

karlpauls:
Hm, on second thought I think I agree that this sounds odd. I'll have to look into the it test some more to see if it really captures the right thing because it almost looks like it is not checking enough to be sure it tests what it is supposed to test.

karlpauls:
I think you can close this PR (see SLING-5457 for details).

PR #201, opened 2017-03-06
SLING-6614: check that property/parent node is not protected before import

glucazeau:


rombert:

rombert:
@glucazeau  - sorry, I just noticed that this went unreviewed. If you're still interested, can you please resubmit this against https://github.com/apache/sling-org-apache-sling-jcr-contentloader ?

PR #202, opened 2017-03-14
SLING-6592 introduce "Content" interface instead of nested maps

stefanseifert:


bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/Content.java
```diff
@@ -0,0 +1,52 @@
...
+ * under the License.
+ */
+package org.apache.sling.jcr.contentparser;
+
+import java.util.Map;
+
+/**
+ * Represents a resource or node in the content hierarchy.
+ */
+public interface Content {
```
kwin, line 26: Can't we just rely on org.apache.sling.api.resource.Resource and throw a https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html for the methods which are not useful in this  context, which are
- `getResourceMetadata()`
- `getResourceResolver()`
maybe `isResourceType(String resourceType)`.

kwin (changes requested):

stefanseifert:
closing this one, it's superseded by #203

PR #203, opened 2017-03-16
SLING-6592 switch to stream-based API for ContentParser

stefanseifert:


bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/ContentHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sling.jcr.contentparser;
+
+import java.util.Map;
+
+/**
+ * Handle that gets noticed while parsing content with {@link ContentParser}.
+ * The resources are already reported in order of their paths as found in the content source.
```
kwin: This javadoc should state, whether parents are reported before or after the children.

bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/ContentHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface ContentHandler {
+
+    /**
+     * Resource found in parsed content.
+     * @param path Path of resource inside the content fragment. The root resource from the content fragment has a path "/".
+     * @param properties Resource properties
+     */
+    void resource(String path, Map<String,Object> properties);
```
kwin, line 35: Maybe split `path` into `parentPath` and `name`. That way it would be easier to persist this in a JCR or via Sling CRUD API.

bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/ContentHandler.java
stefanseifert: fixed this one - parents are always reported before their children.

bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/ContentHandler.java
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sling.jcr.contentparser;
+
+import java.util.Map;
+
+/**
+ * Handler that gets noticed while parsing content with {@link ContentParser}.
```
kwin: noticed -> notified

bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/ContentHandler.java
stefanseifert: i want to leave it as it is to have simple an clean API here. splitting path in parentPah and name is trivial operation if the consumer needs it.

bundles/jcr/contentparser/src/main/java/org/apache/sling/jcr/contentparser/ContentHandler.java
stefanseifert: javadocs updated

stefanseifert:
its applied to svn

PR #204, opened 2017-03-16
Feature/sling 6652

buuhuu:
This pull requests enables 1 to many binding from resourceType to model.

IMHO the implementation is quite ugly because of the way resourceType binding is defined. In the current interface the untyped methods `getModelFromResource(Resource)` and `getModelFromRequest(SlingHttpServletRequest)` have been added to `ModelFactory`. Those are not conform to the way the `ModelFactory` implementation work: creating a typed object from something. I understood that the usecase for the jackson exporter doesn't require a typed object as exporting an ordinary one works quite well but it makes hard to decide which adapterType should be used in case multiple are registered for a single resourceType. Having said that I deprecated those methods along with related ones, but still supporting their expected functionality, with only a slightly difference as adapterTypes are now sorted alphabetically - the same way its done for the original adapter implementations. Anyway removing those, or throwing `USOE` would make sense, to make the API clean again. Most of the code added to `AdapterImplementations` can and should be moved to the related `ImplementationPicker` using it which is supposed to decide which implementation to use based on the adaptable's resourceType. For the exporter a slightly adjustment would be necessary, which is now also implemented: Keeping in mind that class level annotations are not inherited, there is a 1:1 relation between `@Exporter` and its implementation class and with that we can implicitly add the `implType` to the `adapterTypes` if not already present. As the `implType` is known to the `ExporterServlet` anyway it can use it to get the typed object for exporting.

### DIFF
--- a/bundles/extensions/models/api/src/main/java/org/apache/sling/models/factory/ModelFactory.java
+++ b/bundles/extensions/models/api/src/main/java/org/apache/sling/models/factory/ModelFactory.java
@@ -50,7 +50,6 @@ public interface ModelFactory {
             InvalidAdaptableException, ModelClassException, PostConstructException, ValidationException, InvalidModelException;
 
     /**
-     *
      * @param adaptable the adaptable to check
      * @param type the class to check
      * @return {@code true} in case the given class can be created from the given adaptable, otherwise {@code false}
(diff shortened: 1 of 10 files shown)


justinedelson:
Please do not include formatting changes with functional pull requests.

As I commented on https://issues.apache.org/jira/browse/SLING-6655, the methods you are deprecating are used by scripting implementations. They should not be deprecated (or removed).

PR #206, opened 2017-03-17
SLING-6658:

buuhuu:
- always reigster the model with its implType, even not specified as adapter explicitly


kwin:
Can you prepare a patch for https://svn.apache.org/repos/asf/sling/site/trunk/content/documentation/bundles/models.mdtext as well which explains this change (and in the best case links this JIRA issue)?

PR #209, opened 2017-03-31
[contrib esx] removed org.apache.sling.commons.json dependencies and lceaned up the demo" esx/

labertasch:
[contrib esx] removed org.apache.sling.commons.json dependencies and updated the demo


labertasch:
linked in jira ticket: https://issues.apache.org/jira/browse/SLING-6763

PR #212, opened 2017-04-11
SLING-6763 |  fixed typo in readme for the test script foo.esx and up…

labertasch:
SLING-6763 |  fixed typo in readme for the test script foo.esx and updated parent pom to 30


rombert:
@labertasch - the change has been applied, can you please close this PR?

PR #213, opened 2017-04-12
SLING-6780 - org.apache.sling.scripting.sightly.js.provider does not …

vladbailescu:
…declare an version range for org.mozilla.javascript import

* Added dependency to org.apache.sling.scripting.javascript

### DIFF
--- a/bundles/scripting/sightly/js-use-provider/pom.xml
+++ b/bundles/scripting/sightly/js-use-provider/pom.xml
@@ -123,6 +123,12 @@
         </dependency>
         <dependency>
             <groupId>org.apache.sling</groupId>
+            <artifactId>org.apache.sling.scripting.javascript</artifactId>
+            <version>2.0.30</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.sling</groupId>
             <artifactId>org.apache.sling.serviceusermapper</artifactId>
             <version>1.2.0</version>
             <scope>provided</scope>
--- a/bundles/scripting/sightly/js-use-provider/pom.xml
+++ b/bundles/scripting/sightly/js-use-provider/pom.xml
@@ -70,6 +70,7 @@
                     <instructions>
                         <Sling-Bundle-Resources>/libs/sling/sightly/js;path:=/SLING-INF/libs/sling/sightly/js</Sling-Bundle-Resources>
                         <Require-Capability>io.sightly; filter:="(&amp;(version&gt;=1.0)(!(version&gt;=2.0)))"</Require-Capability>
+                        <Import-Package>org.mozilla.javascript;version="[1.1,2)",*</Import-Package>
                     </instructions>
                 </configuration>
             </plugin>
@@ -119,12 +120,6 @@
             <groupId>org.apache.sling</groupId>
             <artifactId>org.apache.sling.scripting.core</artifactId>
             <version>2.0.32</version>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.sling</groupId>
-            <artifactId>org.apache.sling.scripting.javascript</artifactId>
-            <version>2.0.30</version>
             <scope>provided</scope>
         </dependency>
         <dependency>


bundles/scripting/sightly/js-use-provider/pom.xml
```diff
@@ -70,6 +70,7 @@
                     <instructions>
                         <Sling-Bundle-Resources>/libs/sling/sightly/js;path:=/SLING-INF/libs/sling/sightly/js</Sling-Bundle-Resources>
                         <Require-Capability>io.sightly; filter:="(&amp;(version&gt;=1.0)(!(version&gt;=2.0)))"</Require-Capability>
+                        <Import-Package>org.mozilla.javascript;version="[1.1,2)",*</Import-Package>
```
kwin, line 73: Can you add a comment while relying on the automatically generated headers from bnd is not enough here?

vladbailescu: Because the import for `org.mozilla.javascript` is not resolved via a bundle (it's resolved directly via `org.mozilla.rhino-1.7R4`) the bnd plugin will not generate a version range for the import. If you also have `org.mozilla.javascript` version 2+ on the same Sling instance, the package import will then resolve to the higher/incompatible package.

kwin: What I meant was adding a comment in the pom.xml (not in this PR), so that one understands even later why this was necessary. Not sure I completely follow you, because what you are saying is that the dependency towards  rhino (https://github.com/vladbailescu/sling/blob/0ec63ac17191b8b2ad839924d39c11e32435192e/bundles/scripting/sightly/js-use-provider/pom.xml#L174) does not generate a version in the import which is correct. But IMHO a bundle should only have bundle dependencies unless it packages the dependency into its own bundle. Therefore I would rather recommend to switch to another dependency (i.e. the bundle which exports Rhino in the correct version).

vladbailescu: Since SLING-6668 `o.a.s.scripting.javascript` is no longer embedding Rhino. @raducotescu and I think it's best we defined the version range manually rather than through the servicemix bundle (which BTW was resulting in `[1.7,2)` range, which is more restrictive than necessary)

vladbailescu:
Patched by @raducotescu in https://svn.apache.org/viewvc?view=revision&revision=1791224

bundles/scripting/sightly/js-use-provider/pom.xml
raducotescu: @kwin, I'll add the comment in the `pom.xml` file.

The thing is that the JS Use Provider was using the same Rhino version as the `org.apache.sling.scripting.javascript` bundle was embedding until recently. However, the provider was not depending on the JavaScript bundle, which meant that the `org.mozilla.javascript` package was imported without an import range.

Since SLING-6668, the `org.mozilla.javascript` package is exported by the `servicemix` bundle that @oliverlietz brought into the Sling launchpad. However, that exports version 1.7, while the JS Provider works ok with anything in the [1.1, 2) range, hence the manually defined import range.

kwin: I understand, but do not agree here.

> ...which BTW was resulting in [1.7,2) range, which is more restrictive than necessary

This is actually the correct range. You depend on Rhino 1.7R4 therefore most probably this bundle will not run on an older Rhino version (like 1.6).

raducotescu: The more I look at this issue, the more I tend to think that the `org.mozilla.javascript;version=1.1.0` export from `org.apache.sling.scripting.javascript` bundle was wrong...

@kwin, @vladbailescu, should we actually define the range as `[1.7, 2)`, given that both  `org.apache.sling.scripting.javascript` and the JS Use Provider relied on Rhino 1.7?

kwin: Manually maintaining version ranges very soon becomes a mess. Please only depend on non-bundle dependencies if you embed those. Otherwise depend on the bundles and rely on bnd to generate the correct version ranges.

PR #218, opened 2017-04-25
sling-3270 minimal extension of the anchor tag

JEBailey:
allows the anchor tag to wrap additional content. this will address the most common issue being reported with using tagsoup and html5


JEBailey:
I've had some conversations, I'm working on an improved status

PR #222, opened 2017-05-03
SLING-6826 i18n: Supported nested keys in JSON i18n files

stefanseifert:


stefanseifert:
ticket closed as "won't fix" - see SLING-6826 for details

PR #223, opened 2017-05-06
SLING-6506 - Tracer - Fixed query/plan correlation when union'd queri…

davidjgonzalez:
…es are followed by 1+ queries

Resolves: https://issues.apache.org/jira/browse/SLING-6506

Example output post patch (addressing erring use-case in SLING-6505 ticket)

image


rombert:

PR #224, opened 2017-05-09
Fix a cross-site scripting vulnerability in the SlingLogPanel.

chrisgavin:
The log appender name is a user provided value originating [here](/apache/sling/blob/f0e4561935cbebfe6f84a85622513baa74e72d44/bundles/commons/log-webconsole/src/main/java/org/apache/sling/commons/log/webconsole/internal/LogWebConsolePlugin.java#L65), therefore it should not be written to the output HTML without being sanitized.


h3xspirit:
Which versions are exactly affected? We are using 2.11.0. Thank you

chrisgavin:
From what I can tell, this has been present since mid-2015. I'm not sure what versions that means it affects though.

rombert:

PR #225, opened 2017-05-11
Adds missing assignment of name in MockRequestParameter.

Obirah:
I stumbled across this missing assignment when I wrote unit tests for Adobe AEM in combination with AEM mocks (wcm.io.testing).


rombert:

PR #228, opened 2017-05-13
SLING-6855 Add ResultRegistry

cgoudie:
I want to provide a Registry service that can be leveraged to provide health check results.
These results can be for a period of time through an expiration, until the JVM is restarted, or added and later removed.
This can be useful when code observes a specific (possibly bad) state, and wants to alert through the health check API that this state has taken place.
Some examples: 
An event pool has filled, and some events will be thrown away.
This is a failure case that requires a restart of the instance.
It would be appropriate to trigger a permanent failure.
A quota has been tripped. This quota may immediately recover, but it is sensible to alert for 30 minutes that the quota has been tripped.
If you expect the failure will clear itself within a certain window, setting the expiration to that window can be ideal.

### DIFF
--- a/bundles/extensions/healthcheck/api/pom.xml
+++ b/bundles/extensions/healthcheck/api/pom.xml
@@ -67,6 +67,12 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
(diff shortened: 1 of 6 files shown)


rombert:

cgoudie:
@bdelacretaz Should we rescue this or leave this feature to die?

PR #231, opened 2017-05-17
SLING-6516 Annotation Based ValidationModelProvider evaluating additional annotations on a Sling Model bound to a specific resource type and registering ValidationModels

karolisma:
Annotation Based ValidationModelProvider evaluating additional annotations on a Sling Model bound to a specific resource type and registering ValidationModels


bundles/extensions/validation/api/src/main/java/org/apache/sling/validation/annotations/Validate.java
```diff
@@ -0,0 +1,63 @@
...
+     * @return the string
+     */
+    String validatorId() default "";
+
+    /**
+     * Properties array to be passed for validator.
+     *
+     * @return validator properties
+     */
+    String[] properties() default {};
```
kwin: Since properties is a term which is used a lot, I would prefer naming it arguments. Also the javadoc can maybe something like "Parametrisation being passed to the validator given by the id. Each string entry must have the format `<key>=<value>`"

bundles/extensions/validation/api/src/main/java/org/apache/sling/validation/annotations/Validate.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The Validate annotation to be used on Sling Model class fields to provide extra validation arguments.
+ */
+@Target({ METHOD, FIELD, PARAMETER })
+@Retention(RUNTIME)
+public @interface Validate {
+
+    /**
+     * Validator id, with which it is registered.
```
kwin: Rather "Validate the annotated element with a Validator having the given id".

bundles/extensions/validation/api/src/main/java/org/apache/sling/validation/annotations/Validate.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return the string
+     */
+    String validatorId() default "";
```
kwin: I am wondering why this can be empty. IMHO this annotation only makes sense in case the id is set as well, except for the case with the regex.

bundles/extensions/validation/api/src/main/java/org/apache/sling/validation/annotations/Validate.java
```diff
@@ -0,0 +1,63 @@
...
+     * @return the Severity
+     */
+    int severity() default 0;
+
+    /**
+     * Regex string.
+     *
+     * @return regex
+     */
+    String regex() default "";
```
kwin: Please clarify the javadoc here. Also since this can be used independent of an actual validator I would propose to rather come up with a dedicated annotation or not supporting this in the first place for Sling models (don't have a good use case in mind).

bundles/extensions/validation/api/src/main/java/org/apache/sling/validation/annotations/ValidationPaths.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * The Validation paths annotation to be used on Sling Model class to provide applicable validation paths.
+ */
+@Target(TYPE)
+@Retention(RUNTIME)
+public @interface ValidationPaths {
```
kwin: I think ApplicableValidationPaths is a clearer name.

bundles/extensions/validation/core/pom.xml
```diff
@@ -171,6 +171,12 @@
             <version>2.2.0</version>
             <scope>provided</scope>
         </dependency>
+        <!-- For @Model Validation https://issues.apache.org/jira/browse/SLING-6516 -->
+        <dependency>
+            <groupId>org.apache.sling</groupId>
+            <artifactId>org.apache.sling.models.api</artifactId>
+            <version>1.3.3-SNAPSHOT</version>
```
kwin: why does it depend on the newest snapshot? Please reference here the lowest models API you can work with.

bundles/extensions/validation/core/src/main/java/org/apache/sling/validation/impl/annotationmodel/AnnotationValidationModelProviderImpl.java
```diff
@@ -0,0 +1,72 @@
...
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * AnnotationValidationModelProviderImpl provides Validation Models based on Annotations.
+ * It is responsible for creation and registering of ValidationPackageBundleListener.
+ */
+@Component(immediate = true, service = ValidationModelProvider.class)
```
kwin: No need for immediate=true, it will be automatically activated once referenced.

bundles/extensions/validation/core/src/main/java/org/apache/sling/validation/impl/annotationmodel/ValidationModelRegister.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
+
+import javax.annotation.Nonnull;
+
+import org.apache.sling.validation.model.ValidationModel;
+import org.osgi.framework.Bundle;
+
+/** The Validation model register keeps all registered validation models. */
+final class ValidationModelRegister {
```
kwin: rather ValidationModelRegistry

bundles/extensions/validation/core/src/main/java/org/apache/sling/validation/impl/annotationmodel/AnnotationValidationModelProviderImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Activate.
+     * Registers validation package bundle listener.
+     *
+     * @param ctx the component context
+     */
+    @Activate
+    protected void activate(final ComponentContext ctx) {
+        this.listener = new ValidationPackageBundleListener(ctx.getBundleContext(), this.validationModelRegister);
```
kwin: How do you initially fill the registry with bundles which have been active before this bundle?

bundles/extensions/validation/core/src/main/java/org/apache/sling/validation/impl/annotationmodel/ValidationPackageBundleListener.java
```diff
@@ -0,0 +1,192 @@
...
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.BundleEvent;
+import org.osgi.util.tracker.BundleTracker;
+import org.osgi.util.tracker.BundleTrackerCustomizer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** The Validation package bundle listener. It analyzes all the bundles upon their activation, finds Sling Models marked with
+ * ValidationStrategy.REQUIRED or ValidationStrategy.OPTIONAL and registers/unregisters validation models. */
+public class ValidationPackageBundleListener implements BundleTrackerCustomizer {
```
kwin, line 48: I am wondering whether we can reuse some of the logic already contained in Sling Models Bundle listener? What about the former registers an event listener which is triggered by event being emitted from sling Models whenever a new model class is added/removed.

kwin (changes requested):

rombert:

PR #232, opened 2017-05-18

nitin-nizhawan:
- Interpret parsed restriction clauses from repoinit parser

### DIFF
--- a/bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
+++ b/bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
@@ -17,15 +17,20 @@
 package org.apache.sling.jcr.repoinit.impl;
 
 import java.security.Principal;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.jcr.PathNotFoundException;
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
 import javax.jcr.UnsupportedRepositoryOperationException;
+import javax.jcr.Value;
+import javax.jcr.ValueFactory;
 import javax.jcr.security.AccessControlEntry;
 import javax.jcr.security.AccessControlManager;
 import javax.jcr.security.Privilege;
@@ -35,7 +40,11 @@ import org.apache.jackrabbit.api.security.JackrabbitAccessControlList;
 import org.apache.jackrabbit.api.security.JackrabbitAccessControlManager;
 import org.apache.jackrabbit.api.security.user.Authorizable;
 import org.apache.jackrabbit.commons.jackrabbit.authorization.AccessControlUtils;
+
 import org.apache.jackrabbit.oak.spi.security.principal.EveryonePrincipal;
+
+import org.apache.sling.repoinit.parser.operations.RestrictionClause;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -43,8 +52,7 @@ import org.slf4j.LoggerFactory;
 public class AclUtil {
 
     private static final Logger LOG = LoggerFactory.getLogger(AclUtil.class);
-    
-    public static JackrabbitAccessControlManager getJACM(Session s) throws UnsupportedRepositoryOperationException, RepositoryException {
+    public static JackrabbitAccessControlManager getJACM(Session s) throws RepositoryException {
         final AccessControlManager acm = s.getAccessControlManager();
         if(!(acm instanceof JackrabbitAccessControlManager)) {
             throw new IllegalStateException(
@@ -54,7 +62,66 @@ public class AclUtil {
         return (JackrabbitAccessControlManager) acm;
     }
 
+    /**
+     * checks if provided restriction is supported at
+     * current path or not
+     *
+     * @param specifiedRestrictions
+     * @param jacl
+     * @return
+     */
+    private static void validateRestrictionClauses(List<RestrictionClause> specifiedRestrictions, JackrabbitAccessControlList jacl) throws RepositoryException{
+        String [] supportedRestrictionNames = jacl.getRestrictionNames();
+        Set<String> supportedRestrictionsSet = new HashSet<>(Arrays.asList(supportedRestrictionNames));
+        for(RestrictionClause rc : specifiedRestrictions){
+            if(!supportedRestrictionsSet.contains(rc.getName())) {
+                throw new IllegalStateException("Unsupported restriction "+rc.getName()+" at "+jacl.getPath());
+            }
+        }
+    }
+
+    /**
+     * Converts RestrictionClauses to structure consumable by
+     * jackrabbit
+     * @param list
+     * @param jacl
+     * @param s
+     * @return
+     * @throws RepositoryException
+     */
+    private static LocalRestrictions createOakRestrictions(List<RestrictionClause> list, JackrabbitAccessControlList jacl, Session s) throws RepositoryException {
+        Map<String,Value> restrictions = new HashMap<String,Value>();
+        Map<String,Value[]> mvrestrictions = new HashMap<String,Value[]>();
+
+        if(list != null && !list.isEmpty()){
+            ValueFactory vf = s.getValueFactory();
+
+           for(RestrictionClause rc : list){
+               String restrictionName = rc.getName();
+               int type = jacl.getRestrictionType(restrictionName);
+               Value[] values = new Value[rc.getValues().size()];
+               for(int i=0;i<values.length;i++) {
+                   values[i] = vf.createValue(rc.getValues().get(i),type);
+               }
+
+               if(values.length > 1) {
+                   mvrestrictions.put(restrictionName, values);
+               } else {
+                   restrictions.put(restrictionName, values[0]);
+               }
+           }
+        }
+        return new LocalRestrictions(restrictions,mvrestrictions);
+
+    }
+
+
     public static void setAcl(Session session, List<String> principals, List<String> paths, List<String> privileges, boolean isAllow)
+            throws UnsupportedRepositoryOperationException, RepositoryException {
+        setAcl(session,principals,paths,privileges,isAllow,Arrays.asList(new RestrictionClause[]{}));
+    }
+
+    public static void setAcl(Session session, List<String> principals, List<String> paths, List<String> privileges, boolean isAllow, List<RestrictionClause> restrictionClauses)
             throws UnsupportedRepositoryOperationException, RepositoryException {
 
         final String [] privArray = privileges.toArray(new String[privileges.size()]);
@@ -64,8 +131,15 @@ public class AclUtil {
             if(!session.nodeExists(path)) {
                 throw new PathNotFoundException("Cannot set ACL on non-existent path " + path);
             }
+
             JackrabbitAccessControlList acl = AccessControlUtils.getAccessControlList(session, path);
+
+            validateRestrictionClauses(restrictionClauses, acl);
+
+            LocalRestrictions oakRestrictions = createOakRestrictions(restrictionClauses,acl,session);
+
             AccessControlEntry[] existingAces = acl.getAccessControlEntries();
+
             boolean changed = false;
             for (String name : principals) {
                 final Principal principal;
@@ -81,12 +155,13 @@ public class AclUtil {
                 if (principal == null) {
                     throw new IllegalStateException("Principal not found: " + name);
                 }
-                LocalAccessControlEntry newAce = new LocalAccessControlEntry(principal, jcrPriv, isAllow);
+                LocalAccessControlEntry newAce = new LocalAccessControlEntry(principal, jcrPriv, isAllow, oakRestrictions);
                 if (contains(existingAces, newAce)) {
                     LOG.info("Not adding {} to path {} since an equivalent access control entry already exists", newAce, path);
                     continue;
                 }
-                acl.addEntry(newAce.principal, newAce.privileges, newAce.isAllow);
+                acl.addEntry(newAce.principal, newAce.privileges, newAce.isAllow,
+                             newAce.oakRestrictions.getRestrictions(),newAce.oakRestrictions.getMVRestrictions());
                 changed = true;
             }
             if ( changed ) {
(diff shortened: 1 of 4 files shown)


bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -54,8 +62,67 @@ public static JackrabbitAccessControlManager getJACM(Session s) throws Unsupport
...
+
+           for(RestrictionClause rc : list){
+               String restrictionName = rc.getName();
+               int type = jacl.getRestrictionType(restrictionName);
+               Value[] values = new Value[rc.getValues().size()];
+               for(int i=0;i<values.length;i++) {
+                   values[i] = vf.createValue(rc.getValues().get(i),type);
+               }
+
+               if(values.length > 1) {
```
anchela, line 107: the length of the value array of a mv-restriction could potentially be zero... so i think checking for values > 1 is a bug and will lead to indexoutofboundsexception for the empty array. i would suggest to change it to values.length == 1

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -54,8 +62,67 @@ public static JackrabbitAccessControlManager getJACM(Session s) throws Unsupport
...
+    }
+
+
     public static void setAcl(Session session, List<String> principals, List<String> paths, List<String> privileges, boolean isAllow)
             throws UnsupportedRepositoryOperationException, RepositoryException {
+        setAcl(session,principals,paths,privileges,isAllow,Arrays.asList(new RestrictionClause[]{}));
+    }
+
+    public static void setAcl(Session session, List<String> principals, List<String> paths, List<String> privileges, boolean isAllow, List<RestrictionClause> restrictionClauses)
+            throws UnsupportedRepositoryOperationException, RepositoryException {
```
anchela, line 125: declaring _UnsupportedRepositoryOperationException_ is redundant here... in the static 'getJACM' above you removed it (unrelated change). so i would suggest to drop it here as well for consistency

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -54,8 +62,67 @@ public static JackrabbitAccessControlManager getJACM(Session s) throws Unsupport
...
+               }
+           }
+        }
+        return new LocalRestrictions(restrictions,mvrestrictions);
+
+    }
+
+
     public static void setAcl(Session session, List<String> principals, List<String> paths, List<String> privileges, boolean isAllow)
             throws UnsupportedRepositoryOperationException, RepositoryException {
```
anchela, line 120: declaring UnsupportedRepositoryOperationException is redundant here... in the static 'getJACM' above you removed it (unrelated change). so i would suggest to drop it here as well for consistency

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -64,8 +131,15 @@ public static void setAcl(Session session, List<String> principals, List<String>
             if(!session.nodeExists(path)) {
                 throw new PathNotFoundException("Cannot set ACL on non-existent path " + path);
             }
+
             JackrabbitAccessControlList acl = AccessControlUtils.getAccessControlList(session, path);
+
+            validateRestrictionClauses(restrictionClauses, acl);
+
+            LocalRestrictions oakRestrictions = createOakRestrictions(restrictionClauses,acl,session);
```
anchela, line 139: why are they called 'oakRestrictions'? that looks a bit confusing to me as the restrictions are part of the jackrabbit api... i would drop the mentioning of oak here unless you are not making use of Oak API/SPI.

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -81,12 +155,13 @@ public static void setAcl(Session session, List<String> principals, List<String>
                 if (principal == null) {
                     throw new IllegalStateException("Principal not found: " + name);
                 }
-                LocalAccessControlEntry newAce = new LocalAccessControlEntry(principal, jcrPriv, isAllow);
+                LocalAccessControlEntry newAce = new LocalAccessControlEntry(principal, jcrPriv, isAllow, oakRestrictions);
```
anchela, line 158: same here...  i don't think that referring to oak for the restrictions is really correct here, is it?

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -81,12 +155,13 @@ public static void setAcl(Session session, List<String> principals, List<String>
...
                 if (contains(existingAces, newAce)) {
                     LOG.info("Not adding {} to path {} since an equivalent access control entry already exists", newAce, path);
                     continue;
                 }
-                acl.addEntry(newAce.principal, newAce.privileges, newAce.isAllow);
+                acl.addEntry(newAce.principal, newAce.privileges, newAce.isAllow,
+                             newAce.oakRestrictions.getRestrictions(),newAce.oakRestrictions.getMVRestrictions());
```
anchela, line 164: same here

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -122,21 +197,60 @@ private static String toString(JackrabbitAccessControlEntry entry) throws Reposi
...
         private final boolean isAllow;
-        
+        private final LocalRestrictions oakRestrictions;
+
         LocalAccessControlEntry(Principal principal, Privilege[] privileges, boolean isAllow) {
+            this(principal, privileges, isAllow, null);
+        }
+
+        LocalAccessControlEntry(Principal principal, Privilege[] privileges,
+                                boolean isAllow, LocalRestrictions oakRestrictions) {
```
anchela, line 207: and here.
btw: what does 'local' stands for? and local to what?

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -122,21 +197,60 @@ private static String toString(JackrabbitAccessControlEntry entry) throws Reposi
...
             this.principal = principal;
             this.privileges = privileges;
             this.isAllow = isAllow;
+            this.oakRestrictions = oakRestrictions != null ? oakRestrictions : new LocalRestrictions();
```
anchela, line 211: same again

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -122,21 +197,60 @@ private static String toString(JackrabbitAccessControlEntry entry) throws Reposi
...
+                        }
+                    }
+                }
+                return true;
+            }
+            return false;
+        }
+
         private boolean contains(Privilege[] first, Privilege[] second) {
             // we need to ensure that the privilege order is not taken into account, so we use sets
```
anchela, line 255: that's not your fault... but this method has a major bug as it doesn't take into account that there are aggregated privileges. a set that just compares privileges won't work.... this method needs to take the aggregation into account if it wants to be correct.

a simple test would have revealed this: take privilege jcr:read and test if rep:readNodes is contained. or alternatively: jcr:modifyProperities or jcr:write (which is even defined an aggregated in the JCR specification) or jcr:all (the most obvious).

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -54,8 +62,67 @@ public static JackrabbitAccessControlManager getJACM(Session s) throws Unsupport
...
 
+    /**
+     * checks if provided restriction is supported at
+     * current path or not
+     *
+     * @param specifiedRestrictions
+     * @param jacl
+     * @return
+     */
+    private static void validateRestrictionClauses(List<RestrictionClause> specifiedRestrictions, JackrabbitAccessControlList jacl) throws RepositoryException{
```
anchela, line 73: not sure if that is really needed... IMHO the implementation of the Oak RestrictionProvider is in charge of validating the restrictions latest upon save. i would rely on that instead of duplicating the validation, which IMO goes beyond just checking for supported names. in other words: a given restriction definition may still fail upon save if some other constraint is violated (e.g. passing wrong type or passing mv when single value is expected or if a mandatory restriction is missing).

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -54,8 +62,67 @@ public static JackrabbitAccessControlManager getJACM(Session s) throws Unsupport
...
+     * Converts RestrictionClauses to structure consumable by
+     * jackrabbit
+     * @param list
+     * @param jacl
+     * @param s
+     * @return
+     * @throws RepositoryException
+     */
+    private static LocalRestrictions createOakRestrictions(List<RestrictionClause> list, JackrabbitAccessControlList jacl, Session s) throws RepositoryException {
+        Map<String,Value> restrictions = new HashMap<String,Value>();
```
anchela, line 93: i would drop 'oak' from the method name. first of all you are not creating Oak restrictions but only relying on Jackrabbit API and second you are effectively returning an "LocalRestriction", which in turn has no reference to Oak restriction API... so no reason to mention oak here. I would just call it "createRestrictions" or "createLocalRestrictions", which is what you are really doing, right?

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
```diff
@@ -122,21 +197,60 @@ private static String toString(JackrabbitAccessControlEntry entry) throws Reposi
         private final Principal principal;
         private final Privilege[] privileges;
         private final boolean isAllow;
-        
+        private final LocalRestrictions oakRestrictions;
```
anchela, line 200: and here again :-) these are not oak restrictions

bundles/jcr/repoinit/src/test/java/org/apache/sling/jcr/repoinit/GeneralAclTest.java
```diff
@@ -400,4 +440,132 @@ public void mergePreserveMode_PreserveAceTest() throws Exception {
...
+        }
+    }
+
+
+    /**
+     * Tests use of rep:itemNames restriction in set ACL
+     * @throws Exception
+     */
+    @Test
+    public void itemNamesRestrictionTest() throws Exception {
```
anchela, line 485: i would suggest to also create a test with empty-mv- restriction (see above for the minor bug wrt this)

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Hi, @anchela I agree with you that this is a bug.  Issue here is that we do not know at this point whether the restriction is mv or not. So, we are creating mv restriction if there are more than 1 values.  Would you know, if jackrabbit has an API to find if a restriction is mv?

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Ok, makes sense

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Ok, I will remove it in that case

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: I think it got added after merging upstream changes, Anyway I will remove it

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: sure

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: will remove it

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: will remove it

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Will rename it to "localRestrictions"

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Local means local to this class (inspired byLocalAccessControEntry :)  ) , just to separate from "Restriction"

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: 👍  I will log a bug for this since it is unrelated to this change. Thanks for the test case

bundles/jcr/repoinit/src/test/java/org/apache/sling/jcr/repoinit/GeneralAclTest.java
nitin-nizhawan: ok

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
anchela: @nitin-nizhawan unfortunately that's still missing in the jackrabbit API... but my point wasn't primarily about the case when you have 1 value. it's about an empty value array. as long as there is no way to find out about the mv-status, i would test for value length being 1 (and relying on RestrictionProvider impls to deal with a single value if the target is multivalued).

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Logged JIRA for this https://issues.apache.org/jira/browse/SLING-6867

bundles/jcr/repoinit/src/main/java/org/apache/sling/jcr/repoinit/impl/AclUtil.java
nitin-nizhawan: Oh thanks, @anchela I completely missed ioobe. I will update this.

nitin-nizhawan:
Closing this pull request, will raise fresh pr

PR #234, opened 2017-05-21

nitin-nizhawan:
- Interpret parsed restriction clauses from repoinit


rombert:

PR #236, opened 2017-05-30
Feature/slingbucks2

schaefa:
Rewrite of the Slingbucks sample with HTL, service user and JCR Content Package using the new WCMIO Content Package Plugin (1.5.0).

For the time being the WCMIO Content Package must be download from https://github.com/wcm-io/wcm-io-tooling/releases/tag/wcmio-content-package-maven-plugin-1.5.0 and installed locally until the plugin is available on Maven Central.


rombert:

Note that we'll likely revisit the package situation once the filevault-package-maven-plugin has its first release in Jackrabbit.

PR #237, opened 2017-05-30
Feature/sling project archetypes

schaefa:
Created two archetypes with a service core and a content package ui.apps. The 'project' is deployed through the ui.apps module whereas the 'project-all' is using a wrapper content package which contains the core and ui.apps.

Both archetypes come with shadow folders both for the core and ui.apps with example code. The code is generated in a way that it can be easily copied over to the active module but if not needed it can be deleted w/o any side effects.


rombert:

Note that we'll likely revisit the archetype situation once the filevault-package-maven-plugin has its first release in Jackrabbit.

PR #246, opened 2017-07-06
SLING-6078 sling pipes fluent API

npeltier:
- allow easy creation & run of pipes from script & java,
- added some tests


npeltier:
(not sure why 8826c6a is included here)

PR #249, opened 2017-08-02
method to stop a bundle added.

AndreeaDieaconu:
public void stopBundle(String symbolicName)


rombert:
Thanks for the contribuition @AndreeaDieaconu  . Can you also add a Sling jira issue at https://issues.apache.org/jira/browse/SLING ?

PR #252, opened 2017-09-29
SLING-7140 Support redirects to URLs provided by the underlying datastore.

ieb:
This pull request contains API and pom changes which is why I am not committing it directly myself.
It should be squashed into a Single commit.
It depends on https://github.com/apache/jackrabbit-oak/pull/70 being merged and OAK 1.6.5 (or 6) being released. It will break a build if Oak 1.6.5-SNAPSHOT does not contain that pull request.

It enables Sling to redirect requests for binaries held in Oak in a S3 DataStore to be served by an appropriately configured CloudFront instance.

For more info and documentation see https://issues.apache.org/jira/browse/SLING-7140 and https://issues.apache.org/jira/browse/OAK-6575


ieb:
This pull request should not be merged at this time. An alternative patch is being developed that does not depend on Oak.

PR #256, opened 2017-10-17
Update old snapshot deps to current releases and snapshots

cschneider:
Dependencies point to old snapshots that are not available anymore.
Updated to release where possible and to newest snapshot where no release was available.

### DIFF
--- a/bundles/extensions/discovery/impl/pom.xml
+++ b/bundles/extensions/discovery/impl/pom.xml
@@ -153,7 +153,7 @@
 		<dependency>
 			<groupId>org.apache.sling</groupId>
 			<artifactId>org.apache.sling.discovery.commons</artifactId>
-			<version>1.0.21-SNAPSHOT</version>
+			<version>1.0.20</version>
             <scope>provided</scope>
 		</dependency>
         <!-- besides including discovery.commons' normal jar above, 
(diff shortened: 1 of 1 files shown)


bundles/extensions/discovery/impl/pom.xml
```diff
@@ -153,7 +153,7 @@
 		<dependency>
 			<groupId>org.apache.sling</groupId>
 			<artifactId>org.apache.sling.discovery.commons</artifactId>
-			<version>1.0.21-SNAPSHOT</version>
+			<version>1.0.23-SNAPSHOT</version>
```
rombert: Have you tried using the latest released version? My gut feeling is that the snapshot dep was used for some API changes which are now included in the released discovery.commons ( @stefan-egli - feel free to comment ).

cschneider: I tried but maven was not able to find a 1.0.21 or 1.0.22 release

rombert: Hm, it seems we have a staging repository for 1.0.22 but no release vote yet. So yes, 1.0.23-SNAPSHOT is the right version for it.

eglistefan: 1.0.22 got cancelled (SLING-6992) due to java 6 not applicable anymore anyway

eglistefan: 1.0.20 might be fine though - not sure what diff it has with trunk though

eglistefan: @rombert re "it seems we have a staging repository for 1.0.22":
where do you see that staging repo? Can't find it on https://repository.apache.org/#stagingRepositories

rombert: @stefan-egli - ah, sorry. That's Discovery Oak, not Discovery Commons.

eglistefan: @cschneider, I would suggest to use 1.0.20 for discovery.commons - it looks fine from a svn-changes point of view

cschneider: Ok .. will try with it and update

PR #257, opened 2017-10-17
Update old snapshot deps to current snapshots

cschneider:


rombert:
I think that since it's a sample it's better to use the latest releases if possible.

cschneider:
Will try with releases and update the pr

cschneider:
The release versions worked but I had to also had to add the hc api as dependency.. seems it is not included transitively anymore.

PR #262, opened 2017-10-18
SLING-7140  Support redirects to URLs provided by the underlying datastore.

ieb:
This pull request does not require any changes in Oak.
It introduces 2 interfaces to Sling API to allow a ResourceProvider to implement URIProvider that will allow Sing to redirect to a direct URL on get operations. If no URIProviders are implemented, no redirects are issued. The URIProvider implementation would normally be implemented inside the ResourceProvider eg in the case of a S3DataStore for Oak, in the bundle where that code is implemented.


ieb:
Applied via SVN and into the new Git repo for resource resolver.

